PR #61, opened 2021-01-08
Moved updateConnParams out of onConnect

### LINKED ISSUE #60: Possible connection issue with iPhone
This may or may not come up in your testing but I have found that newer versions of iOS don't seem to like [updateConnParams](https://github.com/doudar/SmartSpin2k/blob/715743afa8b6395a000d38316b4952e424971ce3/src/BLE_Server.cpp#L308) in the server onConnect callback. In my experience lately nRFconnect will fail to connect until that line is removed, it could be due to the parameters I used or it could be a bug in nRFconnect but I thought I'd mention it in case you run into this.

doudar:
Fixes #60

### DIFF
--- a/include/settings.h
+++ b/include/settings.h
@@ -10,7 +10,7 @@
 #define SETTINGS_H
 
 //Current program version info. Used for auto updates
-#define FIRMWARE_VERSION "0.1.1.6"
+#define FIRMWARE_VERSION "0.1.1.7"
 
 //Update firmware on boot?
 #define AUTO_FIRMWARE_UPDATE true
--- a/src/BLE_Server.cpp
+++ b/src/BLE_Server.cpp
@@ -5,19 +5,21 @@
 // This work is licensed under the GNU General Public License v2
 // Prototype hardware build from plans in the SmartSpin2k repository are licensed under Cern Open Hardware Licence version 2 Permissive
 
-
 #include "Main.h"
 #include "BLE_Common.h"
 
 #include <ArduinoJson.h>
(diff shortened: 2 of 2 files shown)


h2zero:
This looks like a patch I would do, aim to fix one thing and change a bunch of stuff at the same time lol.

1 comment though, I don't think you need the delay, but if your experience says otherwise I would find it interesting. 
Weird dynamic with the server side of NimBLE that I'd like to understand.

doudar:
Oh wow, I didn't actually look at the commit :)

That's ugly isn't it. I did a right click - format document in PIO (autoformat). I'll have to keep that in mind. I'm newish to PIO.

The vTaskDelay() isn't probably needed but since you suggested there could be an issue in the callback, I thought it may be timing related so I figured I could spare 100ms.

doudar:
I just make sure to never use autoformat in the files where I define pretty classes :)

Love/hate. Love the auto indents, hate that it mixed up your intentional spaces.

h2zero:
It's all good as long as the code works haha, only an issue if you break something at the same time in the commit :). 

I think the issue came from calling the updateParams within the callback context, however I'm not sure about that. The delay won't hurt but it may not help either.

doudar:
Fwiw I did a bunch more testing tonight and couldn't get it to fail in either spot. 🤷‍♂️

Here's to heading off a potential issue?

h2zero:
Nice, like I said, it was only in my experience and have not heard of others but good to get ahead of it for sure.

PR #63, opened 2021-01-08
Added Github CA Certificate

### LINKED ISSUE #18: OTA Firmware updates
Implement the OTA updates. One issue here is having enough free memory to hold the uploaded firmware but there should be just enough room....

doudar:
fixes #18


doudar (commented):
Looks correct!

PR #67, opened 2021-01-16
Hr to pwr button

doudar:
Added a HTML button to toggle between automatic HR->PWR. 

Added a BLE scan when user selects save on the select bluetooth devices page.  

Default watts reduced from 100 to 30


doudar (commented):
Add that default watts were reduced to 30

PR #71, opened 2021-01-22
Ignore second cadence

doudar:
Lots of changes:

Powermeter will now switch to only the most recently connected.
Double power option in the bluetooth scanner.
Bluetooth scanning now happens via a flag set in the client task because spawning it in the HTTP task was causing crashes. 
Backend of the bluetooth scanning page revamped. 
Removed dedicated server callback for the bluetooth scanner via HTML.


doudar (commented):
Looks Correct!

PR #76, opened 2021-02-06
Telegram

doudar:
Added telegram messaging for BLE development. The device will now send your SSID (no passwords or other sensitive information) and the first few strings of BLE data received from your power meter. I can make this telegram info available as a private group if anyone is interested in seeing it. 

If you compile yourself, you may want to add your own telegram bot info to telegram_token.h . I am unable to include my bot token in the Github file as then anyone would have access to my bot. If you wish to compile this project as your own, you may need to create your own telegram bot (it's very easy) and put the credentials in this file. https://core.telegram.org/bots

In addition, this branch included fixes for a rare BLE bug, added favicon.ico (certain browsers were insistent every website has this), and now all of the builtin links use ip address instead of the MDNS service as some networks don't support MDNS (ie. using SmartSpin2K.local).


doudar (commented):
Looks Correct

PR #84, opened 2021-02-10
Conditional Compile Telegram

MarkusSchneider:
Hi Anthony,
I've added a second build configuration to include telegram.
First build configuration (env.esp32doit) will create a binary without any telegram dependencies.
Second one (env.sep32doit-telegram) creates a binary with telegram support and includes the telelgram-bot library.
I am not using telegram and I maybe will always us the first build configuration.
Maybe not the best solution for a feature toggle but the telegram secrets are in the telegram_token.h and two configurations allow to build binaries with or without telegram support.
An advantage is all so that the binary without telegram support is a little bit smaller  :)
Just take a look into the platform.ini and you see how it will work.

For development it's maybe better to create a feature.h and define there how to compile
e.g #define USE_TELGRAM or not... otherwise in vscode USE_TELEGRAM is not defined an developing will be hard...


doudar:
Awesome! Thanks! I can't wait to take a look at it. 

Have to work my regular job today :)

doudar:
Looks like everything is all good - I'm going to merge it into a new branch, do some quick compile testing and then merge it into the main if all is good (it should be). 

Thanks for the PR!

Issue #87, opened 2021-02-17
Security: Password should not be shown in cleartext


kadaan:
For security, the WIFI password should not be shown in cleartext.

doudar:
Fixed by #88

PR #88, opened 2021-02-17
Hide Wifi password by default

kadaan:
Hide the Wifi password by default and provide an option to show it.

image

image

See: https://github.com/doudar/SmartSpin2k/issues/87


doudar (approved):
Easy enough.

Thanks!!

doudar:
Will merge

doudar:
Fixed #87

PR #99, opened 2021-02-19
Flywheel

doudar:
Doesn't seem to cause issues with CPS or HR so adding it to main. 

The actual use needs testing however


doudar (commented):
Looks Correct

PR #104, opened 2021-02-22
Sensor Data Abstraction

kadaan:
Add sensor data abstraction and implement for Heartrate, Flywheel, and FTMS.

TODO: Cadence and Power Sensor Implementation


doudar (approved):
Impressive!

Can't wait to test it!

kadaan:
@doudar Thanks.  Would love to see Cadence and Power Service fit into this.  That opens up the way to Peloton without mucking up the codebase.

PR #105, opened 2021-02-23
Fitness machine service

doudar:
Initial testing looks great @kadaan I think this can be merged into main for testing into the next release.


src/BLE_Client.cpp
```diff
@@ -212,12 +206,18 @@ bool SpinBLEClient::connectToServer()
             charUUID = FLYWHEEL_UART_TX_UUID;
             debugDirector("trying to connect to Flywheel Bike");
         }
-        else
+        else if (myDevice->isAdvertisingService(CYCLINGPOWERSERVICE_UUID))
```
kadaan, line 209: A later improvement is to pull this code up into the SensorData implementation so that contributing sources does not require changes all over the code

src/BLE_Client.cpp
```diff
@@ -303,7 +303,7 @@ bool SpinBLEClient::connectToServer()
             {
 
                 pRemoteCharacteristic->subscribe(true, notifyCallback);
-                if (pRemoteService->getUUID() == CYCLINGPOWERSERVICE_UUID)
+                if (myDevice == myPowerMeter)
```
kadaan, line 306: @doudar Can you explain these changes so I understand the net effect?

kadaan (commented):
Just a few questions so I can better understand some of the changes.

include/BLE_Common.h
```diff
@@ -36,9 +36,12 @@
 #define FITNESSMACHINEPOWERRANGE_UUID BLEUUID((uint16_t)0x2AD8)
 
 // GATT service/characteristic UUIDs for Flywheel Bike from ptx2/gymnasticon/
-#define FLYWHEEL_UART_SERVICE_UUID BLEUUID("6e400001b5a3f393e0a9e50e24dcca9e")
-#define FLYWHEEL_UART_RX_UUID BLEUUID("6e400002b5a3f393e0a9e50e24dcca9e")
-#define FLYWHEEL_UART_TX_UUID BLEUUID("6e400003b5a3f393e0a9e50e24dcca9e")
+#define FLYWHEEL_UART_SERVICE_UUID BLEUUID((uint16_t)0xCA9E)
+#define FLYWHEEL_UART_RX_UUID BLEUUID((uint16_t)0xCA9E)
+#define FLYWHEEL_UART_TX_UUID BLEUUID((uint16_t)0xCA9E)
+//#define FLYWHEEL_UART_SERVICE_UUID BLEUUID("6e400001b5a3f393e0a9e50e24dcca9e")
```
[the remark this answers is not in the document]
doudar: Only because they are untested in the new form. I'm not familiar enough with the conversion to the short form to know for 100% they are correct until I get a report back from a Flywheel bike owner.

src/BLE_Client.cpp
doudar: Changed because as we keep adding more UUID's we're checking against, the line was getting long and I realized we already had a variable that always contained what we needed to check against for simplification.

doudar: Mydevice being the device we're trying to connect and mypowermeter being the selected powermeter found during the last scan.

kadaan: Does the "selected" one handle "any"?

include/BLE_Common.h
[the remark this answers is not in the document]
kadaan: Got it

src/BLE_Client.cpp
doudar: "Any" or the name of the saved ble server is tested in the client code starting on line 476

kadaan (approved):

PR #107, opened 2021-02-24
Fix mDNS dropouts

kadaan:
Fix mDNS dropouts by adding a service text record that gets updated every minute.

Screen Shot 2021-02-24 at 2 47 18 AM

### DIFF
--- a/src/HTTP_Server_Basic.cpp
+++ b/src/HTTP_Server_Basic.cpp
@@ -100,6 +100,7 @@ void startWifi()
   }
 
   MDNS.addService("http", "_tcp", 80);
+  MDNS.addServiceTxt("http", "_tcp", "lf", "0");
   debugDirector("Connected to " + String(userConfig.getSsid()) + " IP address: " + myIP.toString(), true, true);
   debugDirector(String("Open http://") + userConfig.getDeviceName() + ".local/");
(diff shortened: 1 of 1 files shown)


doudar (commented):
"Add" sounds like we keep appending this key value pair to some sort of array.  Are we sure that running MDNS.addServiceTxt("http", "_tcp", "lf", String(mDnsTimer)); repeatedly every minute doesn't cause a memory leak on the ESP32 or cause possible downstream issues with ledgers on local LAN devices? 

Also, every minute seems excessive. Is it required that often in your testing?

kadaan:
`addServiceText` is an add or set, so in our case it is an update. My understanding was that the mDNS resolution stopped working after 2m, so 1m felt safe to try. I’ve left SS2K running all night so we can see how it is doing.

doudar:
> `addServiceText` is an add or set, so in our case it is an update. My understanding was that the mDNS resolution stopped working after 2m, so 1m felt safe to try. I’ve left SS2K running all night so we can see how it is doing.

If you haven't,  you could add in Serial.println(ESP.getFreeHeap()); to monitor memory usage via the serial terminal for debugging.

doudar (approved):
Moving this into PR-TESTING for ... testing......

PR #119, opened 2021-03-08
Set hostname for dns

kadaan:
https://github.com/espressif/arduino-esp32/issues/2537#issuecomment-590029109 gave a workaround so that the hostname for DNS could be correctly sent.  This will allow accessing the the website by going to `http://smartspin2k/"

image

### DIFF
--- a/data/settings.html
+++ b/data/settings.html
@@ -43,7 +43,7 @@
             </tr>
             <tr>
               <td>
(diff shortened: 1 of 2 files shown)


doudar (commented):
So it looks like this shouldn't interfere with the MDNS service either correct?

kadaan:
> So it looks like this shouldn't interfere with the MDNS service either correct?

Correct

PR #125, opened 2021-03-08
Revert "Set hostname for dns"

kadaan:
Reverts doudar/SmartSpin2k#119

### DIFF
--- a/data/settings.html
+++ b/data/settings.html
@@ -43,7 +43,7 @@
             </tr>
             <tr>
               <td>
-                <p class="tooltip">Hostname<span class="tooltiptext">DNS Name the device will use on the
+                <p class="tooltip">MDNS Name<span class="tooltiptext">DNS Name the device will use on the
                     network.</span></p>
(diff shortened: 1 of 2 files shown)


doudar:
It actually seems like `WiFi.config(INADDR_NONE, INADDR_NONE, INADDR_NONE);` Sets the ipaddress to 255.255.255.255 in binaries that I've compiled. 

I wonder if this issue the commenter referenced has been fixed on later versions of the IDE?

kadaan:
AFAIK this this hadn't been fixed, but we can punt on this and test better  later.

doudar:
> AFAIK this this hadn't been fixed, but we can punt on this and test better later.

I haven't been able to test personally, but I sent a binary to @fireyeti and he said the debug output said it was at 255.255.255.255 and was not accessible. I removed the commit and it is working normally. 

I'm currently using  framework-arduinoespressif32 | framework |   | ~3.10005.0

PR #127, opened 2021-03-09
Reduce instantiation

kadaan:
@doudar Critical look please


doudar (commented):
Might want to add simulatedSpeed to String userParameters::returnJSON() in SmartSpin_Patameters.cpp as that function is used to send userParameters to the webserver. 

I like how you did the speed passthrough.

Everything else looks good. (Except you hit autoformat, but I understand.) 

You put a lot of work into this! I can't wait to test it.

kadaan:
RE: auto format - sorry about that. Guess it would be good to have some sort of autoformat that we are all cool with, so that we can keep all contributions following the same pattern.

kadaan:
Re: speed - something is wrong with the numbers. They were high, but not double. I want to check the raw information I captured and confirm.

doudar:
@kadaan, I think this is ready to merge and I think I've resolved the conflicts. I'll let you pull the trigger.

kadaan:
I am thinking we will want to clear out sensor data factory on a rescan.

doudar:
> RE: auto format - sorry about that. Guess it would be good to have some sort of autoformat that we are all cool with, so that we can keep all contributions following the same pattern.

In a perfect world, I'd like to try to space out classes like I did with userParameters.h . In the real world the $h1T hits the fan debugging and it looks like BLEClient.cpp does right now.

kadaan:
The sensor stuff could all be in different files.

doudar:
> > RE: auto format - sorry about that. Guess it would be good to have some sort of autoformat that we are all cool with, so that we can keep all contributions following the same pattern.
> 
> In a perfect world, I'd like to try to space out classes like I did with userParameters.h . In the real world the $h1T hits the fan debugging and it looks like BLEClient.cpp does right now.

I guess once we're not under so much heavy development I'll be able to make things look pretty.

kadaan:
One think to think about, the cache key in sensordatafactory is the blechar uuid. Is this unique per device?  If not, it might need to be. Otherwise, not sure how CPS will work with two different pedals.

doudar:
> One think to think about, the cache key in sensordatafactory is the blechar uuid. Is this unique per device? If not, it might need to be. Otherwise, not sure how CPS will work with two different pedals.

I don't think it is. It's just a regular uint_8 I think. You could combine it with the client connection ID. (low int that doesn't change unless the client gets deleted)

doudar:
> > One think to think about, the cache key in sensordatafactory is the blechar uuid. Is this unique per device? If not, it might need to be. Otherwise, not sure how CPS will work with two different pedals.
> 
> I don't think it is. It's just a regular uint_8 I think. You could combine it with the client connection ID. (low int that doesn't change unless the client gets deleted)

Currently `void SpinBLEClient::removeDuplicates(BLEClient *pClient)` gets invoked if the client detects two power sources. It's a workaround for the two power meter issue and I'm good with it that way until the next release.

doudar:
CPS Cadence rarely getting decoded. I'll take a look. I increased the loop time to 1000ms so maybe that's the issue but that's what it was in previous versions so it shouldn't be causing a problem:

image

PR #131, opened 2021-03-13
Adding reference photos for Joroto X2

bau-sec:
Added photos w/ measurements for the Joroto X2


doudar (approved):
Looks Great! 

Thank you!

PR #141, opened 2021-03-19
Add files via upload

KR-Riker:
Hello All,

Doudar originaly mentioned this as a possibility for the SS2KND (Narrow Design). This suggestion is to move the circuit for resisitor "up" and eliminate the pin holes. The case would then have a fence printed (or milled) at the bottom similar to the stepper motor (lid) fence, which would allow the PCB to be "plugged in" (glue optional).

-keith


doudar:
> Hello,
> I see you have been busy with SS2K. I thought I would try out GitHub. I don't really expect any change, just wanted to have some content "to post". My head is spinning. LOL Not really sure what I am doing here. ha ha
> -keith

Nice to see you on here Keith! 

It seems you are looking to reduce the total length of the PCB? 

I think the direction this needs to go is to get rid of both of the existing "daughter boards" and have pcbway fab our own custom control board (open source) that has a low voltage switching DC-DC converter (30vmax->3.3v), ESP32(IC and antenna only), TMC2225(IC only) & a USB interface for the ESP32. This shouldn't be a big design problem for someone good with PCB design (you know anyone?) as most of it would be just copying the schematics from the ESP32 Devkit & TMC2225, both of which are widely available schematics.

 https://dl.espressif.com/dl/schematics/esp32_devkitc_v4-sch.pdf

https://www.trinamic.com/fileadmin/assets/Products/Eval_Drawings/TMC2209_SilentStepStick_Schematic_V20.pdf

doudar:
Then we could decide on a form factor that might be a "one size fits all" for version 3 of the PCB

KR-Riker:
Yea, shorter PCB could be desirable,  however the ESP-32 doesn't appear to be "shortenable".  Although with short SS2K PCB, the ESP-32 as is could then be "plugged/mounted into a pocket or fence (vertically).  A shorter (and maybe slightly wider - 50x45mm +/-) combo might provide the ability to mount PCB semi-horizontally in the space above the big gear/pulley.

My physics/electronics colleague is a "roller blader" soo.... (not likely Swiftable but a good guy) then there is COVID,  maybe by Fall 2021, I can run it by him.  LOL  That is the only part of your big comment I could comprehend.  Good to hear from you, though.

PR #142, opened 2021-03-20
Change FitnessMachineFeature to enum for readability

kadaan:


doudar:
I guess my only concerns are what effect this has on the binary size & are we going to be able to apply this uniformly across all of our server characteristics? 

Format looks nice and it seems like it should open up the ability to have dynamic ftms features based on what our client is connected to for remote servers.

doudar (approved):
Conflicts resolved. Merging

PR #144, opened 2021-03-20
Fitness machine feature enum v2

kadaan:
@doudar Try this!


doudar (approved):
👍

doudar:
Works, and the binary size didn't change. Showing 69.7% (used 1371221 bytes from 1966080 bytes)

Party Time!

PR #145, opened 2021-03-23
Flywheel tweaks

ptx2:


doudar (approved):
This looks awesome! 

All reviewed and I don't see any conflicts. 

Thanks for the help - I can't wait until we have time to implement our buffer to get the rest working. 

@kaaden , I'll let you merge this as I'm not certain if we want it going to master or develop. It would be an easy choice normally but we're in this rut again where develop is behind so it's a lot of file changes if I were to try to merge this in to develop.

kadaan:
@ptx2 Thanks!

src/BLE_Client.cpp
```diff
@@ -399,7 +399,7 @@ void SpinBLEClient::MyAdvertisedDeviceCallback::onResult(BLEAdvertisedDevice *ad
     {
         aDevName = "";
     }
-    if ((advertisedDevice->haveServiceUUID()) && (advertisedDevice->isAdvertisingService(CYCLINGPOWERSERVICE_UUID) || advertisedDevice->isAdvertisingService(FLYWHEEL_UART_SERVICE_UUID) || advertisedDevice->isAdvertisingService(FITNESSMACHINESERVICE_UUID) || advertisedDevice->isAdvertisingService(HEARTSERVICE_UUID) || advertisedDevice->isAdvertisingService(ECHELON_DEVICE_UUID)))
+    if ((advertisedDevice->haveServiceUUID()) && (advertisedDevice->isAdvertisingService(CYCLINGPOWERSERVICE_UUID) || (advertisedDevice->isAdvertisingService(FLYWHEEL_UART_SERVICE_UUID) && aDevName == FLYWHEEL_BLE_NAME) || advertisedDevice->isAdvertisingService(FITNESSMACHINESERVICE_UUID) || advertisedDevice->isAdvertisingService(HEARTSERVICE_UUID) || advertisedDevice->isAdvertisingService(ECHELON_DEVICE_UUID)))
```
kadaan, line 402: Geez, this is getting nasty.  We need to solve this quick...

src/sensors/FlywheelData.cpp
```diff
@@ -28,8 +28,8 @@ void FlywheelData::decode(uint8_t *data, size_t length)
 {
     if (data[0] == 0xFF)
     {
-        cadence = float(bytes_to_u16(data[4], data[3]));
-        power = data[12];
+        power = bytes_to_u16(data[3], data[4]); // uint16 big-endian at ofs 3
```
kadaan, line 31: Interesting...  Why is flywheel data so odd...

kadaan (commented):
@ptx2 I assume this "works on your machine" and get you to the point where the data is overwritten too quickly for SS2K to act upon?

ptx2:
> @ptx2 I assume this "works on your machine" and get you to the point where the data is overwritten too quickly for SS2K to act upon?

Exactly! I should note that the power/cadence decoding change is untested (because of the race) but I think it should be right.

src/BLE_Client.cpp
doudar: To only get a lot worse quickly if we keep focusing on the BLE multiplexer aspect, which I'd like to.

doudar:
After the latest merge on develop, this is now an easy fit into Develop. 

Party Time! 🥳

doudar:
@kadaan , THis is ready to merge into Develop. I'm uncertain how to incorporate a new changlog into an existing PR though.

doudar (approved):

PR #157, opened 2021-03-27
Fixing stuck cadence for Assimoa Favero Uno

bau-sec:
Cadence would get stuck at a value close to when I stopped pedaling; it would never go to zero. At the end of an erg mode workout this was particularly annoying because while I was getting off the bike it would just keep increasing resistance because it thought I was still pedaling and it wanted to keep me at a certain wattage.

My setup was using the Assioma Favero Uno pedals. Initially I thought this was due to some `float` weirdness, but the ultimate fix was `bool CyclePowerData::hasCadence() { return !isnan(this->cadence); }` having it match the way `EchelonData.cpp` implemented the method. Basically, `hasCadence()` came back `false` so cadence never got updated when it was zero.

I left my changes in the `CyclePowerData` class (`float` -> `uint16_t`) because these values are being compared to incoming data to look for changes, so it's better to leave them as uint16_t rather than convert everything to floats unnecessarily and then compare.


kadaan:
@bau-sec Thanks so much for contributing 👍   One of our biggest goals is to create a vibrant community around SmartSpin2k!

I'll review this and check it on my end, but since I don't have power pedals I can't test it.  @doudar Can you test this on your end?

kadaan:
@bau-sec Not a requirement, but, if you check the test folder you will see at least one test for FTMS.  At some point it would be good to get some tests for this code in there.  At the moment those tests only run on an ESP32, but my goal is to split out all code that doesn't require ESP32/arduino specific calls into a library and allow testing that can run locally.

src/sensors/CyclePowerData.cpp
```diff
@@ -62,11 +57,13 @@ void CyclePowerData::decode(uint8_t *data, size_t length) {
...
     this->lastCrankRev       = this->crankRev;
-    this->crankRev           = bytes_to_int(data[cPos + 1], data[cPos]);
+    this->crankRev           = bytes_to_u16(data[cPos + 1], data[cPos]);
     this->lastCrankEventTime = this->crankEventTime;
-    this->crankEventTime     = bytes_to_int(data[cPos + 3], data[cPos + 2]);
-    if (this->crankRev > this->lastCrankRev && this->crankEventTime - this->lastCrankEventTime != 0) {
-      int cadence = ((abs(this->crankRev - this->lastCrankRev) * 1024) / abs(this->crankEventTime - this->lastCrankEventTime)) * 60;
+    this->crankEventTime     = bytes_to_u16(data[cPos + 3], data[cPos + 2]);
+    if (this->crankRev > this->lastCrankRev && this->crankEventTime != this->lastCrankEventTime) {
+      float crankChange = abs(this->crankRev - this->lastCrankRev) * 1024;
```
kadaan, line 64: Might be worth marking `crankChange` and `timeElapsed` as `const`.  Thoughts?

src/sensors/CyclePowerData.cpp
```diff
@@ -10,20 +10,15 @@
 
 bool CyclePowerData::hasHeartRate() { return false; }
 
-bool CyclePowerData::hasCadence() { return this->cadence > 0 || this->missedReadingCount < 3; }
+bool CyclePowerData::hasCadence() { return !isnan(this->cadence); }
 
-bool CyclePowerData::hasPower() { return this->power != INT_MIN; }
+bool CyclePowerData::hasPower() { return true; }
```
kadaan, line 15: Why change this to indicate that there is power before a `decode` pass?

src/sensors/CyclePowerData.cpp
```diff
@@ -10,20 +10,15 @@
...
 
 int CyclePowerData::getHeartRate() { return INT_MIN; }
 
-float CyclePowerData::getCadence() {
-  if (!this->hasCadence()) {
-    return NAN;
-  }
-  return this->cadence;
-}
+float CyclePowerData::getCadence() { return this->cadence; }
```
kadaan, line 21: This seems to be fine because we have a requirement that you need to call the `has*` guard before accessing the data.

src/sensors/CyclePowerData.cpp
```diff
@@ -10,20 +10,15 @@
 
 bool CyclePowerData::hasHeartRate() { return false; }
 
-bool CyclePowerData::hasCadence() { return this->cadence > 0 || this->missedReadingCount < 3; }
+bool CyclePowerData::hasCadence() { return !isnan(this->cadence); }
```
kadaan, line 13: Logically, I think this is good.  The value starts as NAN and once we see data it will be set to >= 0.  `missedReadingCount` will still prevent dropouts.

kadaan (changes requested):

doudar:
> Cadence would get stuck at a value close to when I stopped pedaling; it would never go to zero. At the end of an erg mode workout this was particularly annoying because while I was getting off the bike it would just keep increasing resistance because it thought I was still pedaling and it wanted to keep me at a certain wattage.
> 
> My setup was using the Assioma Favero Uno pedals. Initially I thought this was due to some `float` weirdness, but the ultimate fix was `bool CyclePowerData::hasCadence() { return !isnan(this->cadence); }` having it match the way `EchelonData.cpp` implemented the method. Basically, `hasCadence()` came back `false` so cadence never got updated when it was zero.
> 
> I left my changes in the `CyclePowerData` class (`float` -> `uint16_t`) because these values are being compared to incoming data to look for changes, so it's better to leave them as uint16_t rather than convert everything to floats unnecessarily and then compare.

Cool - thanks for the PR.  

This must have been a bug that slipped into the works when we swapped to the new client code.

include/BLE_Common.h
```diff
@@ -49,8 +49,8 @@
 
 // macros to convert different types of bytes into int The naming here sucks and
 // should be fixed.
-#define bytes_to_s16(MSB, LSB) (((signed int)((signed char)MSB))) << 8 | (((signed char)LSB))
-#define bytes_to_u16(MSB, LSB) (((signed int)((signed char)MSB))) << 8 | (((unsigned char)LSB))
+#define bytes_to_s16(MSB, LSB) (((int16_t)((signed char)MSB))) << 8 | (((signed char)LSB))
+#define bytes_to_u16(MSB, LSB) (((uint16_t)((signed char)MSB))) << 8 | (((unsigned char)LSB))
```
doudar, line 53: Good catch - This looks correct but since it's a significant change (signed int to uint16), is anyone able to confirm we don't accidently do weird things around the 128's . Specifically greater than 7.6% grade or >768 watts.

src/sensors/CyclePowerData.cpp
bau-sec: No objection there

src/sensors/CyclePowerData.cpp
bau-sec: To be honest, I was just matching `EchelonData.cpp`. Should `EchelonData.cpp` do the same, or is there a different situation for that one?

include/BLE_Common.h
bau-sec: Maybe just change all `bytes_to_u16` to use `bytes_to_int`? That effectively leaves them unchanged, less testing worry that way?

kadaan: I like that from a testing standpoint. Otherwise we should write some tests to cover the change.

doudar: > Maybe just change all `bytes_to_u16` to use `bytes_to_int` (leaving `CyclePowerData.cpp` using `bytes_to_u16` obviously)? That effectively leaves them unchanged, less testing worry that way?

I like your idea of changing it to uint and int where required. It makes the rest of the code and what's happening more clear.

doudar:
I'll give this a real world test on both FTMS and CPS. Building and riding now.

doudar (approved):
I had a couple small changes during testing. 

Merging. 

Thanks @bau-sec !

include/BLE_Common.h
doudar: I had to revert this because it was throwing impossible numbers in my testing.

doudar (approved):

src/sensors/CyclePowerData.cpp
doudar: @kadaan , feel free to make this change and merge.

kadaan (approved):

PR #161, opened 2021-03-29
Power Correction Factor

kadaan:
Change DoublePower to PowerCorrectionFactor (0.0 - 2.0), allowing correction for bikes (IC4) which broadcast inflated power.


doudar (approved):
Looks appropriate. Thanks!

PR #162, opened 2021-03-29
Added ERGSensitivity setting

### LINKED ISSUE #97: After submitting settings, return to settings page
After submitting settings changes, the result page should notify of success and then return to the settings page.

### LINKED ISSUE #153: [FR] ERG "Aggressiveness" setting
**Is your feature request related to a problem? Please describe.**
Depending on individual bike and rider factors, ERG may take too long or be too aggressive in it's response. 

**Describe the solution you'd like**
There should be a range slider in the settings web page that allows the user to adjust ERG mode aggressiveness.

**Describe alternatives you've considered**
A watts generated/stepper steps table specific to the installation could be dynamically created and saved to spiffs. 

**Additional context**
@bau-sec #149.

doudar:
- Added ERGSensitivity setting in settings.html.
- Added ERGSensitivy parameter.
- Settings page submit returns setting page.
- Added explanatory comments in settings for various values.
- Decreased expected watts per shift from 50 to 30.
Fixes #153 
Fixes #97

Co-Authored-By: bau-sec <6752645+bau-sec@users.noreply.github.com>
Co-Authored-By: Joel Baranick <kadaan@users.noreply.github.com>


src/BLE_Server.cpp
```diff
@@ -142,13 +142,13 @@ void computeERG(int currentWatts, int setPoint) {
...
     // 30  is amount of watts per shift. Was 50, seemed like too much...
-    if (abs(currentWatts - setPoint) < 30) {
-      amountToChangeIncline = (currentWatts - setPoint) * 1;
+    if (abs(currentWatts - setPoint) < WATTS_PER_SHIFT) {
+      amountToChangeIncline = (currentWatts - setPoint) * userConfig.getERGSensitivity();
     }
-    if (abs(currentWatts - setPoint) > 30) {
-      amountToChangeIncline = amountToChangeIncline + ((currentWatts - setPoint)) * 1;
+    if (abs(currentWatts - setPoint) > WATTS_PER_SHIFT) {
+      amountToChangeIncline = amountToChangeIncline + ((currentWatts - setPoint)) * userConfig.getERGSensitivity();
```
bau-sec, line 149: At this point `amountToChange` is `0` so whether the watt change is greater or less than `WATTS_PER_SHIFT`, `amountToChangeIncline` is the same.

src/BLE_Server.cpp
```diff
@@ -142,13 +142,13 @@ void computeERG(int currentWatts, int setPoint) {
...
     }
-    amountToChangeIncline = amountToChangeIncline / ((currentWatts / 100) + 1);
+    amountToChangeIncline = amountToChangeIncline / ((currentWatts / 100) + 1);  // +1 to eliminate possible divide by zero.
 
     // limit to 4 shifts at a time
     if (abs(amountToChangeIncline) > userConfig.getShiftStep() * 5) {
```
bau-sec, line 154: Here `amountToChangeIncline` has units of watt change percentage (I think, might be reading 151 wrong) and we're comparing it to microsteps. Does it need to get scaled by microsteps/shiftstep before this comparison?

src/BLE_Server.cpp
```diff
@@ -142,13 +142,13 @@ void computeERG(int currentWatts, int setPoint) {
 
   if (cad > 20) {
     // 30  is amount of watts per shift. Was 50, seemed like too much...
-    if (abs(currentWatts - setPoint) < 30) {
-      amountToChangeIncline = (currentWatts - setPoint) * 1;
+    if (abs(currentWatts - setPoint) < WATTS_PER_SHIFT) {
+      amountToChangeIncline = (currentWatts - setPoint) * userConfig.getERGSensitivity();
```
bau-sec, line 146: The ultimate change in `Main.cpp`'s `moveStepper()` function is still scaled by incline sensitivity. Should there be a flag that `moveStepper` can see to ignore inclineSensitivity in *erg mode* it it's only affected by the erg sensitivity setting?

src/BLE_Server.cpp
doudar: Good point. I'll add a >=

bau-sec: Here's how I might rework that fragment to make it a little clearer:
```c
void computeERG(int currentWatts, int setPoint) {
  float incline             = userConfig.getIncline();
  int cad                   = userConfig.getSimulatedCad();
  int newIncline            = incline;
  int amountToChangeIncline = 0;
  const int wattChange      = currentWatts - setPoint;
  const float subShiftScale = 1.0; // Can tweak this value lower as you ride to see what's smoothest
  
  if (cad <= 20) {
    // Cadence too low, nothing to do here
   return;
  }

  amountToChangeIncline = wattChange * userConfig.getERGSensitivity();
  // 30  is amount of watts per shift. Was 50, seemed like too much...
  if (abs(wattChange) < WATTS_PER_SHIFT) {
    amountToChangeIncline *= subShiftScale;
  }
  ...
}
```

src/BLE_Server.cpp
doudar: I think ultimately I'd like to get rid of inclineMultiplier entirely. I believe its redundant with a properly configured shiftStep

src/BLE_Server.cpp
doudar: I'll take that. looks good. I've also reordered the max shift capping...Didn't look right to me. 

Here's what I have currently. 
```void computeERG(int currentWatts, int setPoint) {
  // cooldownTimer--;

  float incline               = userConfig.getIncline();
  int cad                     = userConfig.getSimulatedCad();
  float newIncline            = incline;
  float amountToChangeIncline = 0;
  const int wattChange        = currentWatts - setPoint;
  const float subShiftScale   = 1.0;  // Can tweak this value lower as you ride to see what's smoothest

  if (cad <= 20) {
    // Cadence too low, nothing to do here
    return;
  }
  amountToChangeIncline = wattChange * userConfig.getERGSensitivity();
  // 30  is amount of watts per shift. Was 50, seemed like too much...
  if (abs(wattChange) < WATTS_PER_SHIFT) {
    // As the desired value gets closer, make smaller changes for a smoother experience
    amountToChangeIncline *= subShiftScale;
  }
  // limit to 10 shifts at a time
  if (abs(amountToChangeIncline) > userConfig.getShiftStep() * 10) {
    if (amountToChangeIncline > 0) {
      amountToChangeIncline = userConfig.getShiftStep() * 10;
    }
    if (amountToChangeIncline < 0) {
      amountToChangeIncline = -(userConfig.getShiftStep() * 10);
    }
  }
  // Reduce the amount per loop (don't try to oneshot it.) and reduce the movement the higher the watt target is.
  amountToChangeIncline = amountToChangeIncline / ((currentWatts / 100) + .1);  // +.1 to eliminate possible divide by zero.

  newIncline = incline - amountToChangeIncline;  //  }
  userConfig.setIncline(newIncline);
}

Comment (kadaan) [reply src/BLE_Server.cpp:149]: A couple questions/issues:

1. What is the point of `subShiftScale` if it is set to `1.0` and is not configurable without recompiling? 
2. With a "properly" configured SS2K, the only time `subShiftScale` will come into effect if the requested change is less than 30w, which is a pretty small change to begin with.
3. Comparison of `wattChange` to `WATTS_PER_SHIFT` doesn't account for sensitivity
4. It is a little confusing the for an "increase" the "amountToChangeIncline" is negative until the very end of the function.

File: src/BLE_Server.cpp (hunk 3 above)
Comment (kadaan) [reply src/BLE_Server.cpp:146]: Just playing around, felt like it was clear to have ERG compute as multiples of a shiftStep:
```
float amountToShift = (targetWatts - currentWatt) / WATTS_PER_SHIFT * userConfig.getERGSensitivity();
if (abs(amountToShift) > userConfig.getERGMaxShift()) {
  amountToShift = amountToShift / abs(amountToShift) * userConfig.getERGMaxShift();
}
userConfig.setIncline(incline + amountToShift * WATTS_PER_SHIFT);
```

## Assuming
```
WPS=30 ERG_SENS=0.9 ERG_MAXS=1.0 
```

## Example 1
```
CW=100 TW=200 
WC = TW-CW => 100
A2S = WC / WPS ==> 3.33
A2S = A2S * ERG_SENS ==> 2.99
IF ABS(A2S) > 1 THEN A2S = 1 ==> A2S = A2S / ABS(A2S) ==> 1
INCLINE = INCLINE + A2S * WPS ==> 30
```

## Example 2
```
CW=200 TW=100 
WC = TW-CW => -100
A2S = WC / WPS ==> -3.33
A2S = A2S * ERG_SENS ==> -2.99
IF ABS(A2S) > 1 THEN A2S = 1 ==> A2S = A2S / ABS(A2S) ==> -1
INCLINE = INCLINE + A2S * WPS ==> -30
```

## Example 3
```
CW=100 TW=125 
WC = TW-CW => 25
A2S = WC / WPS ==> 0.83
A2S = A2S * ERG_SENS ==> 0.75
IF ABS(A2S) > 1 THEN A2S = 1 ==> A2S = A2S / ABS(A2S) ==> 0.75
INCLINE = INCLINE + A2S * WPS ==> 22.5
```

## Example 4
```
CW=125 TW=100 
WC = TW-CW => -25
A2S = WC / WPS ==> -0.83
A2S = A2S * ERG_SENS ==> -0.75
IF ABS(A2S) > 1 THEN A2S = 1 ==> A2S = A2S / ABS(A2S) ==> -0.75
INCLINE = INCLINE + A2S * WPS ==> -22.5
```

src/BLE_Server.cpp
bau-sec: 1) `subShiftScale` was just giving a name to a magic value that had been there in the past of 0.5. I don't think the goal is to make it user configurable, but I didn't write the original code.
2) My understanding is the goal is to make smaller adjustments when the change is smaller so that a 10W difference causes slower adjustments when closer to the target value. Power reporting can jump up and down a bit so a rider won't be necessarily locked into the target power after hitting it and you don't want to overreact to a small wattage change request.
3) I think that's independent of the sensitivity, it's just asking if the change is bigger or smaller than a shift step

doudar: @bau-sec is correct. The newly termed subShiftScale which only comes into effect close to the target is to subtly adjust close to the target set point. In my recent testing, it wasn't required by the newest tweaks in the function, but it needs to remain because it may become needed again as others test. 

The bane of every ERG users existence is:
* An ERG function which oscillates above and below a set point.
* An ERG function which drastically overshoots targets. 

subShiftScale accounts for the first. 

The second one is accounted for by capping max shifts, not trying to "one shot" the calculation, and decreasing the amount to change as currentWatts increases. It's important for us to not dwell on getting this exact because we need an adjustable one size fits all solution because different bikes have different resistance mechanisms (magnetic vs friction) as well as different methods to actuate those mechanisms (screws vs cantilevered arms).

src/BLE_Server.cpp
doudar: > Getting rid of it sounds good. It's a little confusing to have a setting both in the SS2K and in zwift related to incline difficulty.
> 
> Until then, what about something like this in `moveStepper`:
> 
> ```c
> void moveStepper(void *pvParameters) {
>   int acceleration   = maxStepperSpeed;
>   int targetPosition = 0;
>   int ustepChange = 0; 
> 
>   while (1) {
>     ustepChange = userConfig.getIncline();
>     if (!userConfig.getERGMode()) {
>       // Only use the incline multiplier when not in erg mode
>       ustepChange *= userConfig.getInclineMultiplier());
>     }
>     targetPosition = shifterPosition + ustepChange;
>     ...
> ```
> 
> It's a small change that's easy to get rid of if incline multiplier goes away. The flag already exists so not much effort there. And with this incline multiplier and erg sensitivity are untangled from each other.

Thinking out loud.......

I think some of the variable naming is throwing both you and @kadaan off, and I don't blame you. As the project has evolved, variable usage has subtly changed and possibly outgrown their intended purpose. This project needs more input from users like you guys to help become more accessible to others to modify. 

Maybe it's a naming issue, but (the value of) inclineMultiplier still needs to be incorporated into both modes to fully convert incline units to motor steps whether that incline is internally or externally derived.

To try to help, userConfig.getIncline() returns a target which is set either sim mode or internally via our ERG calc. Our ERG is purely specifying a change to incline to achieve target watts.

If you took into account my bike's normal settings (shiftStep = 800, inclineMultiplier=4), it would appear that (the value of) inclineMultiplier (4) could be derived (at least in my case) by doing (.005*userConfig.getShiftStep()). I think it's reasonable to assume .005 may be some type of constant that we can use to convert incline to motor steps. I'm also fairly certain that the constant of .005 is applicable across the board to all bikes. 

So we can get rid of inclineMultipler exactly that easy. Multiply userConfig.getShiftStep() by .005 . The reason I haven't is because apps other than Zwift may not have a "difficulty slider". The the inclineMultiplier is exactly a difficulty slider in sim mode. 

So we need a difficulty slider in some apps and not others but using this value currently affects our ERG mode (which is undesirable.) I suggest this option:

* Move inclineMultiplier's modification out of the motor loop and into bleServer.cpp's  onWrite() callback for sim mode where we have it modify userConfig.setIncline() directly. 
* Rename inclineMultiplier to "simModeDifficulty"
* Change motor control loop to convert userConfig.getIncline() into steps by doing: 
``const float inclineToStepConversion = userConfig.getShiftSteps() * .005;
targetPosition = shifterPosition + (userConfig.getIncline() * inclineToStepConversion);    ``

Therefore, it really doesn't matter on the stepper control loop which mode we're in. It's just going to convert userConfig.getIncline() to stepper steps and determine if that's a ccw or cw rotation.

doudar: Modifying this however is outside of the scope of this particular PR, and i'll do it on a separate branch.

kadaan: Changing `inclineMultiplier` to `simModeDifficulty` is confusing as it is applicable to bother `Sim` and `ERG` modes.

bau-sec: Regardless of the ultimate ERG/Sim algorithm, we'll probably need a set of unittests that can effectively do what @kadaan did w/ his examples above for different step sizes to make sure the microsteps scale properly as the algorithm gets tuned.

For that you'd want a slight refactor of `computeERG` so it returns the incline value rather than setting it directly. Might need some refactoring of `moveStepper` functionality to make it more unittest-able as well depending on how the algorithm is split between `computeERG`, `computeCSC` and `moveStepper`

PR #167, opened 2021-04-01
Short hr server

doudar:
Just needs testing. Last I checked, it seems to be compatible with everything.


doudar (commented):
Tons of work into this one :)

kadaan (approved):
Pending successful test on hardware, this LGTM

PR #168, opened 2021-04-01
Adding unit tests for CyclePowerData

bau-sec:
I grabbed some data from a short session w/ my Assioma Uno pedals for the unit test data. Fixed a bug in CyclePowerData.cpp where it would think it could report cadence on the first reading.

I know @kadaan was working on native unit tests, I don't think this steps on any of that. The valuable piece here is just sample data.


kadaan:
@bau-sec Love the addition of tests!

kadaan (changes requested):
This is awesome!   Have you don't a hardware test yet to ensure that the CPS still works after these changes?  

I know it is onerous, but until our testing is more robust, can't be too careful!

bau-sec:
> 
> 
> This is awesome! Have you done a hardware test yet to ensure that the CPS still works after these changes?
> 
> I know it is onerous, but until our testing is more robust, can't be too careful!

I haven't done a ride w/ it yet. Probably won't be able to until tomorrow evening.

doudar:
> > This is awesome! Have you done a hardware test yet to ensure that the CPS still works after these changes?
> > I know it is onerous, but until our testing is more robust, can't be too careful!
> 
> I haven't done a ride w/ it yet. Probably won't be able to until tomorrow evening.

You just had to get @kadaan excited about more tests didn't you? :sweat_smile:

doudar (commented):
Looks great pending hardware testing. 

I'm actually okay with an esp32-esp32 hardware test via the BT simulator page on this if you guys want to try it since it's a fairly straight forward change. Would be handy to build in that cadence hook to the webserver for this one. Maybe we should tune up the BTsimulator.html since it's been mothballed for a while.

doudar:
Thanks again for the data @bau-sec . I can't remember what bike you have, but if you have an Echelon, I'd love to see you write a test for it.

PR #171, opened 2021-04-02
Native testing

kadaan:


kadaan:
@bau-sec Automated testing!  My feeling is we should get this in and then put the CPS tests in as native tests.

bau-sec:
> 
> 
> @bau-sec Automated testing! My feeling is we should get this in and then put the CPS tests in as native tests.

That's fine with me, no rush on the CPS PR. The CPS unit tests fixed some things, but they were all minor and/or edge cases, nothing that was affecting my actual rides.

lib/SS2K/src/sensors/FlywheelData.cpp
```diff
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
-#include "BLE_Common.h"
+#include "os/endian.h"
```
doudar, line 8: Endian lib is pretty nice - and small. Good find.

platformio.ini
```diff
@@ -7,16 +7,20 @@
 ;
 ; Please visit documentation for the other options and examples
 ; https://docs.platformio.org/page/projectconf.html
+[platformio]
+default_envs = esp32doit
+
 [common_env_data]
 lib_deps = 
-    ;https://github.com/h2zero/NimBLE-Arduino.git#master //Nimble is tracked and including this will probably cause issues.
```
doudar, line 12: Should probably keep this here? I'm guessing we will eventually add it back in?

doudar: Or is your plan to go completely cross platform dependent?

lib/SS2K/include/Constants.h
```diff
@@ -0,0 +1,20 @@
...
+ * All rights reserved
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#pragma once
+
+#include <NimBLEUUID.h>
+
+#define CYCLINGPOWERMEASUREMENT_UUID NimBLEUUID((uint16_t)0x2A63)
```
doudar, line 12: I'm perfectly okay with moving the UUID's out of BLE common.h but not okay with splitting them up. They all need to be moved here or left as a group over there.

platformio.ini
kadaan: Nope

lib/SS2K/include/Constants.h
kadaan: Okay. I was looking at this as an interim step. I’m fine moving them all here now. I think much of the advertising code moves to the lib, so it can be easily tested.

kadaan:
Still need to ride on SS2K with this change.

doudar (approved):

doudar (approved):

PR #191, opened 2021-04-20
Added mounts, reorganized hardware files.

doudar:
Co-Authored-By: jkbaker4 <40582757+jkbaker4@users.noreply.github.com>
Co-Authored-By: marklogan17 <75054215+marklogan17@users.noreply.github.com>


doudar:
Double check I didn't do something stupid and that the file movements & directory structure makes sense. The bottom case .sldprt deletion was intentional - junk file.

doudar (commented):
Double checked on my end. I'm going to remove the .dxf & .dwg files as they were just from me messing around with different export formats.

PR #192, opened 2021-04-20
Hardware updates

doudar:
- Added more README.md's to /hardware/
- Fixed minor spulling errurs.


doudar (commented):
Quick look by the author looks great :)

fireyeti (approved):
William Shakespeare would be jealous of this prose.

PR #200, opened 2021-04-25
Adjusted Revmaster insert inner diameter

doudar:


doudar:
Old one was 64mm, you measured 61.4mm, I made this 63mm

fireyeti (approved):

PR #217, opened 2021-05-22
Broadcast locally shifted resistance changes over FTMS status characteristic

urb4nc4rl:
Updated to broadcast any hard-triggered local changes to resistance over the FTMS status characteristic.

This means that, if you use the local shifter controls, any remotely connected BLE central devices will receive a notification that the resistance level has changed.

This is achieved via adding an additional flag in Main.cpp `resistanceIsDirty` that is set by the IRAM methods `shiftUp` or `shiftDown` to indicate when the resistance has been changed.

An additional method called by the main run loop, `broadcastChangesIfRequired()` monitors this flag.  If the flag becomes set, it is reset and the method `BLE_Server::updateFTMStatusCharWithResistance()` is called to broadcast the new resistance to any connected centrals.


urb4nc4rl:
@doudar Seems to work well with what I had in mind.  I've added PR #217 to further enhance the functionality for when resistance is changed locally.  Built and works, but only limited testing with writing/reading from the new characteristics, have not given it a full workout.

urb4nc4rl:
Please ignore — this PR was set up to merge into the wrong branch.  Have opened new PR #218

PR #258, opened 2021-11-30
Added UDP Logger

MarkusSchneider:
Added an UPD logger.
Logger sends broadcast messages to the WIFI at Port 10.00

I used Logbert: https://github.com/couchcoding/Logbert to collect the UDP logs. 

SmartSpin2k-UPD-Log
0


doudar:
@MarkusSchneider Looking good! I just want to do a test for radio time sharing before I merge it into develop. 

Any interest on adding UDP log viewing to SmartSpin2kflasher? 

https://github.com/SmartSpin2K/SmartSpin2kFlasher

MarkusSchneider:
Hi Anthony, I've think I can do this. Never used wx but to show UPD Logs in the console window should be possible, like its done with the serial read.

doudar (approved):
Thanks for the hard work @MarkusSchneider !

PR #260, opened 2021-12-01
STEP files and STL's for my ic4 mod.

eMadman:
location in /hardware/mods


doudar (commented):
@eMadman , Looks great!

doudar (approved):

PR #272, opened 2021-12-10
Lots of little changes - see Hardware/MODS/Case V2 - Schwinn IC4 Mod/…

eMadman:
- rename directory to something more apparent
- Removed need for support material from case
- thicker slide design.  Removes need for washers
- Slightly shorter slide - should allow more flexibility for ss2k placement on IC4.
- tighter tolerances for slide fitting.  It should be a snug fit but shouldn't require too much force
- slightly larger diameter holes for m3 screws used in case assembly.  Screws should be much easier to insert
- combined knob cup & knob insert for schwinn to reduce amount of plastic and print time needed


eMadman:
Going to close this pull - @rjbutler  found an issue with my plated stl.

PR #283, opened 2021-12-25
Status.html using rtConfig

MarkusSchneider:
Fixed status.html to use new rtConfig.
configValue only read once if page is loaded, rtConfig is refreshed every 1.5 sec.
If configValues also should be refreshed, put loading of configValues into refresh-timer.

### DIFF
--- a/data/status.html
+++ b/data/status.html
@@ -5,8 +5,8 @@
   <style type="text/css">
     html {
       background-color: #03245c;
-    } 
-        </style>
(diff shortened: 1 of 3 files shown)


doudar:
Looks perfect! This was exactly how I was going to do it. 

Now get back to enjoying your Christmas!

Merry Christmas Markus!

doudar (approved):

PR #287, opened 2021-12-27
Experiments with shifters. Ready to ride

eMadman:


doudar (approved):
We should probably move the old shifter to the Archives folder. I can do that after the PR.

PR #293, opened 2021-12-31
Erg respin

MarkusSchneider:
Hi Anthony,

I've refactored the ERG_mode.
I've moved nearly all code from the task loop into separate methods.
Checked in computeErg if the power value > 10 and created to new methods
- first one makes all calculations when set point changes
- second method is the normal function to compute erg.

We can now try to tune both states of an erg step independently.
Main reason why I've moved all code from task loop is hopefully better test ability.

I tried to write unit test but figured out that this is only for library functions possible.
Like the code of the SSK library. I think this is the way to go.

Happy new year and thank you for all the time an work that you have invested.

Greetings
Markus


doudar:
> Hi Anthony,
> 
> 
> 
> I've refactored the ERG_mode.
> 
> I've moved nearly all code from the task loop into separate methods.
> 
> Checked in computeErg if the power value > 10 and created to new methods
> 
> - first one makes all calculations when set point changes
> 
> - second method is the normal function to compute erg.
> 
> 
> 
> We can now try to tune both states of an erg step independently.
> 
> Main reason why I've moved all code from task loop is hopefully better test ability.
> 
> 
> 
> I tried to write unit test but figured out that this is only for library functions possible.
> 
> Like the code of the SSK library. I think this is the way to go.
> 
> 
> 
> Happy new year and thank you for all the time an work that you have invested.
> 
> 
> 
> Greetings
> 
> Markus
> 
> 

Nice! I'll take a look and merge it when I get some time (probably tomorrow)

Yes, that last ERG needed some refactoring. I was thinking about doing it myself but I'm very happy for your contribution again!

The BLE server really needs a refactor into class SpinBLEServer{}; but I can never seem to find time to get at it. 😅

doudar (commented):
Okay, I was excited about looking at what you did and it all looks great on my phone. Next step is to see if I can build it quick later today.

doudar:
Also if you can, update the changelog.md as that's why the pre-commit is failing.

doudar (approved):

MarkusSchneider:
> Also if you can, update the changelog.md as that's why the pre-commit is failing.

Sorry I forgot to update it. I tried to get precommit to run but without luke.

doudar:
> > Also if you can, update the changelog.md as that's why the pre-commit is failing.
> 
> 
> 
> Sorry I forgot to update it. I tried to get precommit to run but without luke.

No problem! ☺️

Happy New Year!

PR #294, opened 2022-01-01
UI changes

MarkusSchneider:
Created confirm dialog if reset to defaults is clicked.
image

Added file mask (.bin, .html) to upload dialog.
image

Fixed bug with opacity of loading watermark. (loading watermark has no property opacity -> null exception)


doudar (approved):

doudar:
All looks good, but it doesn't work if the SS2K isn't connected to the internet. 

Is event.preventDefault() part of jquery? 

If so, you should be able to point at the jquery.js.gz we have in /data/

MarkusSchneider:
> All looks good but reset to defaults confirmation doesn't work if the SS2K isn't connected to the internet.
> 
> Is event.preventDefault() part of jquery?
> 
> If so, you should be able to point at the jquery.js.gz we have in /data/
> 
> For best stability, it should probably load on a delay as well. I'm traveling the rest of the day, so I probably won't have time to help you on this.
> 
> Thank you!

Hmmm...
preventDefault is standard javascript. Not necessary to us jquery.
I've tested it without connection to internet and it worked. Only problem that I had was the automatic update overwrites the changes.
Can you please check if settings.html is the new one.
- Open settings.html
- Press F12 (chrome)
- Go to Sources Tab in the dev tools
- Check if the loadDefaults() function is present
image

If so, please add a breakpoint on 250 and 251. If you now click Reset to Defaults. Breakpoint should be hit.
If you reject, event.preventDefault() should be called and no form.submit() is send to server.

If it is not working please comment 😉.
I can try to do the form.submit on an other way.

doudar:
> > All looks good but reset to defaults confirmation doesn't work if the SS2K isn't connected to the internet.
> > Is event.preventDefault() part of jquery?
> > If so, you should be able to point at the jquery.js.gz we have in /data/
> > For best stability, it should probably load on a delay as well. I'm traveling the rest of the day, so I probably won't have time to help you on this.
> > Thank you!
> 
> Hmmm... preventDefault is standard javascript. Not necessary to us jquery. I've tested it without connection to internet and it worked. Only problem that I had was the automatic update overwrites the changes. Can you please check if settings.html is the new one.
> 
> * Open settings.html
> * Press F12 (chrome)
> * Go to Sources Tab in the dev tools
> * Check if the loadDefaults() function is present
>   image
> 
> If so, please add a breakpoint on 250 and 251. If you now click Reset to Defaults. Breakpoint should be hit. If you reject, event.preventDefault() should be called and no form.submit() is send to server.
> 
> If it is not working please comment 😉. I can try to do the form.submit in an other way.

So I just did a bunch of testing and it's weird. 

Steps to repeat:

1. On iPhone, go into airplane mode and connect to the SmartSpin2k SSID.
2.  /hotspot-detect.html is served and the portal automatically pops up.
3. Open the SS2K settings webpage and click "Reset to Defaults".
4. event.preventDefault() doesn't fire. 

If I click cancel on the iPhone settings page and "use without internet" go into Safari web browser it works perfectly. 
 
So that leads you to think event.preventDefault() is missing or blocked on the iPhone hotspot portal but If the SS2K is connected to a AP with internet access, the popup works perfectly on the hotspot portal. 

It looks like the iPhone is requesting some other pages we don't have when it tried to do this, so I'll add some additional debugging and see what it's requesting.

PR #303, opened 2022-01-13
Revised shifters, improved printing instructions

eMadman:
@doudar - i had to blow up my local repo, re-clone and push again for these changes.  Can you make sure I haven't made changes that don't belong?  My changes should be to the shifters folder and creating an archive folder for the original design to sit in.


doudar (approved):
It all looks good to me. I don't see any conflicts.

PR #309, opened 2022-01-20
Update readme.md

eMadman:


doudar:
Looks good! Thanks!

Is there anything else downstream in the wiki that we should edit as well?

PR #321, opened 2022-02-03
Added selective notifications

doudar:
- bluetoothscanner.html now lists fitness machine services in the PM list. 
- Updated Arduino_esp32 to the latest 2.0.2 version.
- Fixed all libraries to static releases.
- Reduced max_connect_retries from 10 to 3.
- Increased max_scan_retries from 1 to 2. 
- Now only send notifications for subscribed characteristics.


doudar:
@MarkusSchneider ,
 I might have to roll back to arduino-esp32.git#1.0.8 because of some BLE regression bugs we’re encountering in idf 4.4. 
 
In a test, the only issue I see with rolling back is that your new logging is using <freertos/message_buffer.h> . That library isn’t in the older version of freeRTOS in IDF 3x. 
 
Is it possible to resolve that dependency a different way?

src/BLE_Client.cpp
```diff
@@ -20,20 +20,19 @@ appearance: 1156, manufacturer data: 640302018743, serviceUUID:
...
+static MyClientCallback myClientCallback;
+
 void SpinBLEClient::start() {
   // Create the task for the BLE Client loop
   xTaskCreatePinnedToCore(bleClientTask,   /* Task function. */
                           "BLEClientTask", /* name of task. */
                           4700,            /* Stack size of task */
                           NULL,            /* parameter of the task */
-                          1,               /* priority of the task  */
+                          0,               /* priority of the task  */
```
h2zero, line 35: I don't think you want to change this to idle priority, suggest leaving it as 1

src/BLE_Client.cpp
```diff
@@ -162,116 +167,116 @@ bool SpinBLEClient::connectToServer() {
...
-    } else {  // We don't already have a client that knows this device, we will check for a client that is disconnected that we can use.
-      SS2K_LOG(BLE_CLIENT_LOG_TAG, "No Previous client found");
-      // pClient = NimBLEDevice::getDisconnectedClient();
+    pClient->setClientCallbacks(&myClientCallback, false);
+    /** Set initial connection parameters: These settings are 15ms interval, 0 latency, 120ms timout.
+     *  These settings are safe for 3 clients to connect reliably, can go faster if you have less
+     *  connections. Timeout should be a multiple of the interval, minimum is 100ms.
+     *  Min interval: 12 * 1.25ms = 15, Max interval: 12 * 1.25ms = 15, 0 latency, 51 * 10ms = 510ms timeout
+     */
+    pClient->setConnectionParams(12, 12, 0, 51);
```
h2zero, line 197: Might want to delete this line, the default settings are safer. Test first of course.

doudar: > Might want to delete this line, the default settings are safer. Test first of course.

I'll test it for a while, but I love that it's lightning fast to connect/reconnect, so I'm tempted to keep it.

src/BLE_Client.cpp
doudar: Good call - that's a testing leftover.

src/BLE_Client.cpp
h2zero: As long as you don't get connection issues it should be fine, might want to consider the timeout at least, 510ms is a bit short, maybe set to 100 (1000ms).

doudar: 40 minutes on the test bed with UDP and websocket logging running full blast and multiple simulated disconnects. Everything is perfect so far. 
image

doudar:
> Core 0 is best for BLE tasks, otherwise the IPC is used internally (my little hack to avoid delays). Just an FYI.

I'll try moving them over. It's probably not going to cause an issue anymore. Thanks!

PR #325, opened 2022-02-05
rework of ic4 paths, re-drawn arm and bike mounts, readme revisions

eMadman:
This one's a bit big but it tidies up the schwinn mod into the more generically branded "Direct Mount Mod" and adds Echelon Connect Sport support


doudar (approved):
👍👍

PR #328, opened 2022-02-14
Added a check for wifi connectivity

doudar:
This should fix an issue where if SS2K is plugged in too long, it becomes unreachable. It should now check for wifi every 30 seconds and if it's not connected, it will reconnect.


doudar:
Broken at the moment, the check for WiFi.status() is returning not connected while it still is connected.

PR #342, opened 2022-05-09
LittleFS

doudar:
Swapping in LittleFS instead of (depreciated) SPIFFS. This should fix the occasional settings issues we're experiencing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
  - updated CA for OTA updates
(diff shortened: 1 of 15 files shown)


doudar:
To do:
* Add support for migrating existing settings and files. 
* Feature and Stability Testing.

doudar:
I've had this up and running for a week now and it's stable and responsive. Seems to have much better up time compared to Spiffs. 

Should we merge to develop and then work on how to in-place migrate the user settings later?

eMadman:
@doudar - bench testing has been successful.  I would like to get it onto my bike but haven't been able to make the time. With what I saw in my bench testing, it's proven to be more stable than what's currently in develop.

PR #351, opened 2022-07-08
Release of soldering jig

eMadman:
This has been tested extensively with ABS - no noticeable signs of warping or melting after several PCB's.


doudar (approved):
Looks nice! I love the directions. Nice work!

PR #352, opened 2022-07-08
Added Setting for MinWatts

MarkusSchneider:
Added a setting to define minWatts.
Changed check for minWatts / maxWatts. If setting is 0 minWatts/maxWatts is not set.

image

Range of the minWatts slider is 0 to 200. Steps: 5
Range of maxWatts slider 0 to 2000, steps 25

USERCONFIG_JSON_SIZE was not changed.

I've tried to solve the problem with the duplicated log messages. Only the messages from ERG_Mode log are printed twice and it not always happen. 
I have no idea what the problem is 😒


doudar (approved):
Nice work! The only thing I saw was a ">=" in ERG_Mode.cpp that probably should have been a ">" but you must have fixed it while I was reviewing :)

doudar:
> Added a setting to define minWatts. Changed check for minWatts / maxWatts. If setting is 0 minWatts/maxWatts is not set.
> 
> image
> 
> Range of the minWatts slider is 0 to 200. Steps: 5 Range of maxWatts slider 0 to 2000, steps 25
> 
> USERCONFIG_JSON_SIZE was not changed.
> 
> I've tried to solve the problem with the duplicated log messages. Only the messages from ERG_Mode log are printed twice and it not always happen. I have no idea what the problem is 😒

I didn't notice that it's only ERG mode log messages. That's a place to start looking into this more at least.....

MarkusSchneider:
> I didn't notice that it's only ERG mode log messages. That's a place to start looking into this more at least.....
> 
> Dumb question....Is there any chance we (occasionally at least) have two identical ERG tasks running simultaniously?

I've added a Log Message in the ergTaskLoop Method. 
https://github.com/doudar/SmartSpin2k/blob/a1424c9620b17168e983244c3f68c24682bede8b/src/ERG_Mode.cpp#L32
Only one call was logged...

I've tested with the simulator and received the messages only once. In the debug session without simulator (ERG disabled) I got the message "No data found in powertable." twice.

PR #380, opened 2022-10-05
Simplified CAD compensation in Powertable

doudar:
Binary sent through telegram. Let me know if you think it's ready. I haven't ridden it yet, only tested with the sim.


doudar (commented):
I don't see any mistakes

eMadman:
I'll try to get some on-bike time with this in the next day or two.  Gearing up for Thanksgiving up here

PR #381, opened 2022-10-05
Fixed several bugs in bluetoothscanner.html

doudar:
This should make the bluetoothscanner.html a little more intuitive to use. 

Fixed:

- Instances where many duplicates would be placed in the dropdowns. 
- Currently connected devices are now in the dropdowns and able to be saved. 
- Devices found while scanning will now be a the top of the boxes. 
- The active scan interval has been increased to increase connection chances.


doudar (commented):
I've reviewed the files changed.

alexquilty (commented):
Looks good!

PR #385, opened 2022-10-15
Min max fix

doudar:
@eMadman , leaving this here just so you can review the file changes for your testing. Firmware is attached. Unzip and rename exactly firmware.bin before uploading. 

This hasn't been thoroughly tested and isn't ready to be merged yet. 

[firmware.zip](https://github.com/doudar/SmartSpin2k/files/9792714/firmware.zip)


eMadman:
I'll give it one more test on-bike tomorrow, but I think it's okay to push in the current state.  ERG was bouncing around in my last ride - will try with the erg multiplier down to a lower value

PR #388, opened 2022-10-21
Cad tweaks

eMadman:
- Minor improvements to tolerances for direct mount mod
- created peloton-specific arm for direct mount use.  IC4 model is usable, but a bit short.
- modified short case to include chamfers and fillets at the screw posts to improve thin wall printability in superslicer
- beefier arm for direct mount
- NEW: Direct Mount short case for bikes with reduced clearance in front of knob.  
- NEW: Bolt through short case for direct mount use with Generic Bike http://smile.amazon.com/gp/product/B07S3YWSNM
- NEW: Direct mount for Life Fitness IC7


doudar (approved):
Holy moly! That was a ton of changes @eMadman !!!!!!

PR #391, opened 2022-10-24
Min max fix

doudar:
Fixed a rounding error due to pesky int/int issues.


eMadman (approved):
Test ride with erg mode went great.  Bench test with sim mode is working well.

PR #405, opened 2022-11-27
Auto reconnect

doudar:
This will attempt a scan/connect if no devices are set to "any" and any device NOT set to "none" is disconnected. Interval between scans is ~30 seconds. 

Here's an initial firmware. Not much testing done yet. So far it seems to scan occasionally and not crash :)

As always, extract the firmware.zip, admin/admin and make sure it's named exactly firmware.bin for upload. 

edit: see comments below for the latest build of the firmware from this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added blocking for shifts above or below min/max set points.
 - Added power scaler for new board.
 - Added Main Index link to develop.html.
+- Added feature to automatically reconnect BLE devices if both are specified.
 
 ### Changed
 - PowerTable values are now adjusted to 90 RPM cad on input.
--- a/data/bluetoothscanner.html
+++ b/data/bluetoothscanner.html
@@ -123,7 +123,7 @@
           var t_obj = JSON.parse(data.foundDevices)
           {
             for (var key in t_obj) {
-              if (t_obj[key].UUID == '0x1818' || t_obj[key].UUID == '0x1826') {
+              if (t_obj[key].UUID == '0x1818' || t_obj[key].UUID == '0x1826' || t_obj[key].UUID == '6e400001-b5a3-f393-e0a9-e50e24dcca9e' || t_obj[key].UUID == '0bf669f0-45f2-11e7-9598-0800200c9a66') {
                 optionPM = document.createElement('option');
                 if (t_obj[key].name) {
                   optionPM.text = t_obj[key].name;
--- a/data/bluetoothscanner.html
+++ b/data/bluetoothscanner.html
@@ -152,24 +152,24 @@
           }
         }
 
-        let noneOptionPM = document.createElement('option');
-        noneOptionPM.text = 'none';
-        let noneOptionHR = document.createElement('option');
-        noneOptionHR.text = 'none';
-        PMDropdown.add(noneOptionPM);
-        HRDropdown.add(noneOptionHR);
-        let anyOptionPM = document.createElement('option');
-        anyOptionPM.text = 'any';
-        let anyOptionHR = document.createElement('option');
-        anyOptionHR.text = 'any';
-        PMDropdown.add(anyOptionPM);
-        HRDropdown.add(anyOptionHR);
         let connectedOptionPM = document.createElement('option');
         connectedOptionPM.text = data.connectedPowerMeter;
         let connectedOptionHR = document.createElement('option');
         connectedOptionHR.text = data.connectedHeartMonitor;
         PMDropdown.add(connectedOptionPM);
         HRDropdown.add(connectedOptionHR);
+        let anyOptionPM = document.createElement('option');
+        anyOptionPM.text = 'any';
+        let anyOptionHR = document.createElement('option');
+        anyOptionHR.text = 'any';
+        PMDropdown.add(anyOptionPM);
+        HRDropdown.add(anyOptionHR);
+        let noneOptionPM = document.createElement('option');
+        noneOptionPM.text = 'none';
+        let noneOptionHR = document.createElement('option');
+        noneOptionHR.text = 'none';
+        PMDropdown.add(noneOptionPM);
+        HRDropdown.add(noneOptionHR);
         document.getElementById("connectedPowerMeter").innerHTML = data.connectedPowerMeter;
         document.getElementById("connectedHeartMonitor").innerHTML = data.connectedHeartMonitor;
         document.getElementById("powerCorrectionFactorValue").innerHTML = data.powerCorrectionFactor;
--- a/data/bluetoothscanner.html
+++ b/data/bluetoothscanner.html
@@ -152,18 +152,24 @@
           }
         }
 
-        optionPM = document.createElement('option');
-        optionPM.text = 'none';
-        PMDropdown.add(optionPM);
-        optionHR = document.createElement('option');
-        optionHR.text = 'none';
-        HRDropdown.add(optionHR);
-        let defaultOptionPM = document.createElement('option');
-        defaultOptionPM.text = 'any';
-        let defaultOptionHR = document.createElement('option');
-        defaultOptionHR.text = 'any';
-        PMDropdown.add(defaultOptionPM);
-        HRDropdown.add(defaultOptionHR);
+        let noneOptionPM = document.createElement('option');
+        noneOptionPM.text = 'none';
+        let noneOptionHR = document.createElement('option');
+        noneOptionHR.text = 'none';
+        PMDropdown.add(noneOptionPM);
+        HRDropdown.add(noneOptionHR);
+        let anyOptionPM = document.createElement('option');
+        anyOptionPM.text = 'any';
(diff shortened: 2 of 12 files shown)


doudar:
limited testing - it works pretty good so far. I'll try a regular ride with it here and make sure all is okay, then we should do some exception testing.

doudar:
Working well so far. one potential issue that needs to be addressed is if you have a device configured (i.e. a hr monitor) and the battery were to go dead mid ride, the remaining BLE communication to the client nd server is severely disrupted as it attempts reconnects. 

I think I might be able to find a solution for this.

doudar:
It's better by increasing the connection timeout and decreasing the scan time. The issue is the initial scan needs to be fairly long to catch all devices on the first try. I think I can get it without interruption if I separate the scan times into two different timings. Say a 3 second scan for this method, and a 10 second scan time for other processes. At 3 seconds, we may not catch the device we're looking for on the first scan but this isn't an issue since we automatically re-scan.

doudar:
> It's better by increasing the connection timeout and decreasing the scan time. The issue is the initial scan needs to be fairly long to catch all devices on the first try. I think I can get it without interruption if I separate the scan times into two different timings. Say a 3 second scan for this method, and a 10 second scan time for other processes. At 3 seconds, we may not catch the device we're looking for on the first scan but this isn't an issue since we automatically re-scan.

fixed this. new binary vv
EDIT: See newest binary lower in the comments

eMadman:
Bench test seems to work well.  Test procedure:
Start with connection to a PM, ensure the PM is selected in the BT devices list
Pull power on PM, monitor logs --> confirmed that it's always polling for a new connection

power on PM - SS2K reconnected automatically within ~30 seconds.

eMadman:
Test #2: 

Ride in Mywhoosh with PM & HRM
Disconnect HRM - verify it tries to reconnect every so often
Turn on HRM - verify automatic reconnect

d-nx:
Tested the re-connect feature with my Echelon Sport:

Steps Taken:

1. Set "Select New Power Meter:" to any
2. Save Dropdowns
3. Scan/Connect
4. SS2K connects to Bike
5. Disconnect power to Bike
6. Re-Connect power to Bike, SS2K connects to bike automatically

doudar:
> Tested the re-connect feature with my Echelon Sport:
> 
> 
> 
> Steps Taken:
> 
> 
> 
> 1. Set "Select New Power Meter:" to any
> 
> 2. Save Dropdowns
> 
> 3. Scan/Connect
> 
> 4. SS2K connects to Bike
> 
> 5. Disconnect power to Bike
> 
> 6. Re-Connect power to Bike, SS2K connects to bike automatically
> 
> 
> 
> 

Nice!

Is this ready to be merged into Develop or should we make changes or require further testing?

d-nx:
> > Tested the re-connect feature with my Echelon Sport:
> > Steps Taken:
> > 
> > 1. Set "Select New Power Meter:" to any
> > 2. Save Dropdowns
> > 3. Scan/Connect
> > 4. SS2K connects to Bike
> > 5. Disconnect power to Bike
> > 6. Re-Connect power to Bike, SS2K connects to bike automatically
> 
> Nice!
> 
> Is this ready to be merged into Develop or should we make changes or require further testing?

The only 'weird' thing I noticed was that my bike's name didn't show up under power. Even though it connected, the name didn't show up on the dropdown.

IMO it works. Send it!

doudar:
> > > Tested the re-connect feature with my Echelon Sport:
> > > Steps Taken:
> > > 
> > > 1. Set "Select New Power Meter:" to any
> > > 2. Save Dropdowns
> > > 3. Scan/Connect
> > > 4. SS2K connects to Bike
> > > 5. Disconnect power to Bike
> > > 6. Re-Connect power to Bike, SS2K connects to bike automatically
> > 
> > 
> > Nice!
> > Is this ready to be merged into Develop or should we make changes or require further testing?
> 
> The only 'weird' thing I noticed was that my bike's name didn't show up under power. Even though it connected, the name didn't show up on the dropdown.
> 
> IMO it works. Send it!

Thx! The BLE names are a little different issue I’ll have to fix in the http server. Right now they’re read from only the variable that holds devices found during the (last) scan. If a device is actively connected, it won’t be found during the scan. I think I could fix this by simply adding saved devices to that string before sending. It will occasionally result in duplicates in the list but that’s better than the problem you describe.

doudar:
@d-nx & @eMadman , I fixed bluetooth scanner.html according to the issue by @d-nx . In order to test it, you can manually upload bluetoothscanner.html from this branch via the firmware updater web page. https://github.com/doudar/SmartSpin2k/blob/auto_reconnect/data/bluetoothscanner.html

d-nx:
> @d-nx & @eMadman , I fixed bluetooth scanner.html according to the issue by @d-nx . In order to test it, you can manually upload bluetoothscanner.html from this branch via the firmware updater web page. https://github.com/doudar/SmartSpin2k/blob/auto_reconnect/data/bluetoothscanner.html

I updated the .html file but I'm still seeing the same issue:

Screen Shot 2022-12-02 at 2 48 02 PM

doudar:
Ahh! Thanks for the picture. I remember this issue now. It's because the html is only looking for the cycling power characteristic for PMs. Since you have an Echelon, it's not getting added to the drop-down. I think this is an easy fix. I'll see if I can add it tonight.

doudar:
@d-nx , please let me know if this displays your Echelon after uploading the bluetoothscanner.html. 

You'll probably need to reboot as well to clear any active connections. 

https://github.com/doudar/SmartSpin2k/blob/auto_reconnect/data/bluetoothscanner.html

d-nx:
> @d-nx , please let me know if this displays your Echelon after uploading the bluetoothscanner.html.
> 
> You'll probably need to reboot as well to clear any active connections.
> 
> https://github.com/doudar/SmartSpin2k/blob/auto_reconnect/data/bluetoothscanner.html

It's working a bit better but now I have some duplicates:
Screen Shot 2022-12-02 at 5 37 57 PM

Cleared active connections, rebooted but still see doubles.

d-nx:
Also, it's not allowing me to select 'none' so I can disconnect from the bike.

doudar:
> Also, it's not allowing me to select 'none' so I can disconnect from the bike.

I can't seem to duplicate this - is there a certain browser you're using? I've tried on Windows Chrome, Edge, and IPhone Safari and Chrome. 

As far as the duplicates, I'll have to write a javascript loop in bluetoothscanner.html to remove duplicates in the list. I'm not too concerned about it. 

Super happy the Echelon is showing on the list for you. That was an easy fix.

d-nx:
> > Also, it's not allowing me to select 'none' so I can disconnect from the bike.
> 
> I can't seem to duplicate this - is there a certain browser you're using? I've tried on Windows Chrome, Edge, and IPhone Safari and Chrome.
> 
> As far as the duplicates, I'll have to write a javascript loop in bluetoothscanner.html to remove duplicates in the list. I'm not too concerned about it.
> 
> Super happy the Echelon is showing on the list for you. That was an easy fix.

I'm using Chrome. The behaviour is as follows:

I connect to the bike by selecting "Any" then "Save Dropdowns", after that the bike beeps confirming connection and the name appears on the dropdown menu of SS2K.

Now, I would like to select "None" so the bike disconnects from SS2K (So I can use with a different app), but after hitting "Save Dropdowns" it's still connected. Maybe this has something to do with the new reconnect feature? not sure how to make ss2k disconnect from the bike at this point.

doudar:
Okay, so it will save "none" but the bike doesn't disconnect, correct?

Does it disconnect (and stay disconnected) if you reboot the SS2K after selecting none?

d-nx:
It doesn't save 'none' the echelon remains selected in the dropdown. Ant the bike remains connected to ss2k.

Selecting "None" then "Saving Dropdown" = Echelon remains in the dropdown list, and bike remains connected.
Selecting "None" then "Saving Dropdown" then Rebooting= Echelon remains in the dropdown list, but bike is not connected automatically.

doudar:
> It doesn't save 'none' the echelon remains selected in the dropdown. Ant the bike remains connected to ss2k.
> 
> Selecting "None" then "Saving Dropdown" = Echelon remains in the dropdown list, and bike remains connected. Selecting "None" then "Saving Dropdown" then Rebooting= Echelon remains in the dropdown list, but bike is not connected automatically.

Thanks! I’ll see if I can work on that. I might just have it reboot if none is selected.

doudar:
@d-nx I've changed it so that whenever a BLE device is selected that's different from the currently connected one it will reboot to apply the settings. It's the easiest way to apply the changes that need to be done and it happens pretty fast as long as the WiFi doesn't take a long time to reconnect. 

Newest binary attached. 

[firmware.zip](https://github.com/doudar/SmartSpin2k/files/10171853/firmware.zip)

doudar:
@eMadman see above - any comments?

eMadman:
@doudar device reboots when selecting "None" from the dropdown then quickly finds an available connection.  Seems to be working as intended

eMadman (approved):
Passed testing on bench and bike

PR #406, opened 2022-11-29
echelon insert update

eMadman:
Base was only 1 layer thick - added thickness.


doudar:
I think you should direct this to be merged into "integrated board hardware" as a lot of the paths have changed and it will cause an issue with merging if not.

PR #415, opened 2022-12-17
FTMS Passthrough

doudar:
Allows passthrough of the FTMS data so that an app can control a FTMS device connected through the SS2k


doudar:
No urgent need for merging. It’s a fringe use case that needs to be well tested (especially with ic4).

The concerns are:

Implementation inside isr

Forwarding potentially unsupported commands to the trainer (we’re using the SS2K FTMS capabilities and not the trainers) as there’s no translation (yet)

Increased radio time.

doudar:
Moved this into #420 as they are similar.

PR #420, opened 2022-12-22
Added in resistance reporting for a lot of bikes

doudar:
Added in reading resistance from Echelon, Flywheel and Peloton (and any others that send this data via FTMS) and then outputting that resistance via FTMS server. Needs through testing. Very many files touched.


doudar:
@eMadman - Requesting final review and approval. I think it's ready.

eMadman (approved):
Most recent ride on QZ was great - hitting resistance targets just about instantly.  Zwift is going great as well

PR #432, opened 2023-01-18
Updated PCB

doudar:
Moved Switch Back to board edge
Switched A1 & A2 on Motor Jumper
Rotated back side Silkscreen images
Added Silkscreen to Jumpers on back
updated inductor to in stock model


doudar:
@eMadman - can this be merged?

PR #440, opened 2023-01-23
Disregard Peloton Power IF BLE PM Connected

doudar:
Needs to be tested. 
Need to determine if logging should be reduced with PM data + firehose.


doudar:
@eMadman , please test this and let me know if it can be merged. Thanks!

PR #445, opened 2023-01-29
Added ble Battery level to measurements

Flo100:


doudar:
Goes along with issue #443

I'll work on stability testing and it looks like we should add some changes to the html for displaying this data as well.

Thanks!

doudar:
@Flo100 - Looking good! I could have saved you some work - I thought that characteristic required a subscription, hence adding it into postConnect().

Since it only needs to be read, I just moved it into a function in the SpinBLEClient class and am only calling it from there.

If this works for you, please let me know and then we can continue integrating this so we can access it via the webserver. 

https://github.com/Flo100/SmartSpin2k/blob/67e440296a86fe1d4995900c5763e7684f918cf1/src/BLE_Client.cpp#L623-L647

Flo100:
> @Flo100 - Looking good! I could have saved you some work - I thought that characteristic required a subscription, hence adding it into postConnect().
> 
> Since it only needs to be read, I just moved it into a function in the SpinBLEClient class and am only calling it from there.
> 
No problem. A subscription is also possible. But since it indeed isn't time critical I decided to use polling every 5 minutes.

> If this works for you, please let me know and then we can continue integrating this so we can access it via the webserver.
> 
> https://github.com/Flo100/SmartSpin2k/blob/67e440296a86fe1d4995900c5763e7684f918cf1/src/BLE_Client.cpp#L623-L647

I'll let you know.

I now use it to show it on my Nextion display, along with power, cadence and some other parameters. I had no dropouts or other issues with it so far.
I you're interested I could share the nextion code offcourse.

doudar:
I think it would be awesome if we added a '#define NEXTION' in settings.h and '#ifdef NEXTION' '#endif' around any Nextion code with the main body of Nextion code  in its own .h & .cpp .

It might be a great foundation of down the road supporting displays in the main app. We could always switch to multiple software channels or to a userConfig parameter if you think it would be popular.

doudar:
Also, I would LOVE to see a picture of your setup! Sounds amazing!

Flo100:
> If this works for you, please let me know and then we can continue integrating this so we can access it via the webserver.
> 
> https://github.com/Flo100/SmartSpin2k/blob/67e440296a86fe1d4995900c5763e7684f918cf1/src/BLE_Client.cpp#L623-L647

I get the batterylevel upon connection and it gets updated every 5 minutes. 
It resets to 0 when disconnected, but when the sensor disconnects (and client is reused) the batterylevel does not get updated immediately anymore, but only when the 5 minute timer is past.

doudar:
> > If this works for you, please let me know and then we can continue integrating this so we can access it via the webserver.
> > https://github.com/Flo100/SmartSpin2k/blob/67e440296a86fe1d4995900c5763e7684f918cf1/src/BLE_Client.cpp#L623-L647
> 
> I get the batterylevel upon connection and it gets updated every 5 minutes. It resets to 0 when disconnected, but when the sensor disconnects (and client is reused) the batterylevel does not get updated immediately anymore, but only when the 5 minute timer is past.

Weird that the resets to 0 that were in "onDisconnect" weren't getting called. I placed the same spinBLEClient.handleBattInfo(pClient); function in that callback now. Both of those characteristics should return nullptr on the disconnect so it should read 0. Will see if it fixes it.

Flo100:
Sorry, I see now that I wasn't clear in my reaction and made a typo.
When disconnected it resets to 0 fine. 
The issue occurs when the device REconnects, and the client is reused. In that situation the batterylevel stays 0 until the timer expires. It should update instantly upon reconnection.

doudar:
> Sorry, I see now that I wasn't clear in my reaction and made a typo.
> 
> When disconnected it resets to 0 fine. 
> 
> The issue occurs when the device REconnects, and the client is reused. In that situation the batterylevel stays 0 until the timer expires. It should update instantly upon reconnection.
> 
> 
> 
> 

Okay - I'll take a look at it.

Flo100:
Because handleBattInfo is now a function the last_battery_update counter gets set when a pm OR hrm gets connected the first time. 
When the device disconnects and reconnects the battery update is skipped and we're waiting for the timer to expire.
This is also the case when pm is already connected. The battery info for hrm is skipped upon connection.
I moved it back to onConnect() with an extra parameter to let the function know it should get the batterylevel upon first connect of the device.

PR #470, opened 2023-07-02
Changed wire diameter in the window

doudar:
Wire diameter reduced from 7.2mm to 6.0mm on the window passthrough to accommodate the latest batch of cables.


eMadman (approved):
New window fits fine but it may be recommended to use a zip tie to help retain the cable from behind the window.  The new, narrower cables don't offer as much surface area in contact with the cable for friction alone to do the job.

In daily use, it's fine, but a hard enough tug will pull it loose

PR #480, opened 2023-08-09
Updated arms and IC4 bike mount

eMadman:
tweaks to the openscad model - added roundness.  Increased bolt hole diameter to 5.05mm from 5mm.  Have not test printed 

Can't remember what I changed on the IC4 bike mount but there was a good reason for it.  I've been printing this version for a bit now.


doudar (approved):
Nice find on adding "roundness"!

bannus:
Here's a visual diff of the IC4 mount for posterity. (New is on the right.)

image

doudar:
> Here's a visual diff of the IC4 mount for posterity. (New is on the right.)
> 
> image

The one on the right looks much nicer to get a strap through the slots.

eMadman:
Had one user manage to break the new IC4 mount today.  I I'll try to thicken that part up a touch more before merging.

PR #481, opened 2023-08-09
Fixed Typo in pdf

doudar:
Changed reference to M4 bolt to M5 Bolt.
Increased right side case mounting hole to 5.5mm so the bolt slides in easier.


bannus:
@doudar Any reason not to increase the hole on the left side to 5.5mm as well? It's currently exactly 5mm.

doudar:
> @doudar Any reason not to increase the hole on the left side to 5.5mm as well? It's currently exactly 5mm.

My thinking is that it provides similar locking action to a nylock nut.

doudar:
@eMadman - is this okay to be merged?

eMadman (approved):
Looks good!

Issue #412, opened 2022-12-08
[BUG] WiFi doesn't connect if there is a space in the SSID


doudar:
### Description

SSID doesn't connected if there is a space in the name

### Steps to reproduce

1. Try to connect to a SSID with a space in the name

**Expected behavior**
It should connect

**Actual behavior**
It doesn't connect

### Additional Information
I haven't checked the data yet but I believe this is an issue with how we trim data being sent from the HTML settings page. It's possibly an easy fix.

doudar:
This is fixed by using the ConfigApp

PR #490, opened 2023-11-06
Added a custom integrated implementation of WifiManager.

doudar:
- Updated builtin.h html color scheme.
- When pages are served, BLE server will slow it's roll.
- Added a custom integrated implementation of WifiManager.
- Removed several BLE scan hard stops in favor of a single loop which waits for the scan to gracefully end.
- Fixed bug where WiFi.getMode() wasn't returning the correct mode and caused slow/no wifi connections in AP mode.
- Repurposed LogComData. Will now not transmit any logging info to WiFi when it's turned off.
- Reworked web pages to start to implement more built in functions.
- Removed style.css (Internal to the program now).


eMadman:
this commit should probably come with some refactoring.  I jammed in an individual table per line because I couldn't figure out how to incorporate table rows and cells into HTTP_Server_Basic.cpp.

Also needs testing on multiple OSes, browsers and devices because I didn't apply any styling to the tables beyond an inline align=center

doudar:
Fixes #412

doudar:
This is probably going to be depreciated because of Config app. @eMadman , your thoughts?

eMadman:
Agreed.  The wifi manager and hotspot are  fantastic but they cause
difficulty for some users.  Also bypasses risks associated with users
leaving the device in hotspot mode

PR #498, opened 2023-12-03
Virtual shifting passthrough

doudar:
Added Virtual Shifting for any FTMS trainer as well as implemented a Torque Table for ERG mode instead of the previous Power Table.


doudar:
Now fixes #504

doudar:
Waiting on a successful ride by @eMadman

PR #499, opened 2023-12-03
Yesoul S3

eMadman:
this should do it for the S3.  I think the bike mount could be improved with better measurements.


Hardware/Common Assets/Arm/readme.md
doudar, line 1: Just double checking - Is the 140 the arm length or the actual measurement? That's Yuge!

doudar (commented):
Just check the comment on the arm length.

doudar:
Are these finalized and working?

eMadman:
I just shipped one out recently. I'll check in with the user regarding fit
of the bike mount after it arrives.  I'm pretty sure it's good to go as is
but the fit of the bike mount might not be quite right

eMadman:
Arm length is indeed yuge for the bike

doudar:
> Arm length is indeed yuge for the bike 

Yuge but appropriate, correct?

doudar:
Thanks as usual @eMadman !

PR #505, opened 2023-12-11
Adding bowflex photos and measurements

brooksandrew:
Hi @doudar - we chatted on Etsy earlier today about supporting the Bowflex Velocore.

I'll doublecheck these measurements later this evening, but first wanted to show you the plain images to check if there's any particular measurements you'll need?   I'm measuring with simple tape measure and converting inches to mm.  Is that precise enough?


doudar:
@brooksandrew - Those should mostly work - nice pictures! A couple of notes:

I'll need the width of the oval head tube. Most are 30mm.

Since I can't see the indents in the knob, I'll need you to manually count the grooves in it for me.  
Also, because the straight down top view is shadowing the grooves completely, this means that I'll have to design a split insert  [like this](https://github.com/doudar/SmartSpin2k/blob/develop/Hardware/Common%20Assets/Inserts/Echelon_60mm.STL))

doudar:
Also, a caliper is better, but I've done a few of these with a tape as well. Just get them as close as possible.

brooksandrew:
@doudar - just added some more photos w 1) groove sizing and counts and 2) headtube measurements.   let me know how that looks.  saw some previous photos so included my primitive tape measurement for reference as well.

doudar:
Those are some very nice pictures! Great job!

doudar (approved):
Looks Great! Ready to merge!

PR #546, opened 2024-04-21
Moved wire guard up 1mm

doudar:
Test printing one right now. This is to get rid of an occasional rubbing noise due to interference between the motor shelf and large gear.


doudar:
also fixes #547

doudar:
test print completed and it's good

PR #560, opened 2024-08-01
Amend always option to git describe

jrbarrick:
git_tag_macro.py throws an error "fatal: No names found, cannot describe anything." during PlatformIO initial setup on a new fork.  Looking online, the fix is to add --always to the git describe command.


doudar:
Hi! Thanks for the PR - I don't see that this conflicts with anything, so it should be good. Any chance you can add to the changelog.md in "unreleased" so that the workflow completes?

doudar (approved):

PR #575, opened 2024-09-30
Implemented read and write functionalities for both target watts and if ERG mode is enabled/disabled. 

benv12:
I was able to identify if ERG mode was enabled or disabled by getting the values inside 'FitnessMachineControlPointProcedure'.


doudar (commented):
These all look appropriate. I'll revert the baud changes, just because we have them in a different PR.

doudar (approved):

PR #589, opened 2024-11-07
Added a couple safety features to homing

doudar:
- Added backing off of the stop before we test to prevent runaway grinding during homing. 
- User can abort homing by pressing shifter.


doudar:
We were off by about a shift at the end of the ride

PR #610, opened 2024-12-22
Fix IC bike SE showing only zeroes

### LINKED ISSUE #605: [BUG] Bowflex IC bike SE stops working with latest release.
See discussion https://github.com/doudar/SmartSpin2k/discussions/598

### Steps to reproduce

Flash the latest firmware(24.12.8) and hook smartspin2k to zwift

**Expected behavior**
SmartSpin2K report power and cadence.

**Actual behavior**
SmartSpin2K report zero on power and cadence.

### Additional Information
* https://github.com/doudar/SmartSpin2k/discussions/598 cantains the logs as well for firmwares.
* 24.2.16 works but with some quirks.
* I also found connecting to GT Bike V would crash the game with 24.2.16 not sure if this is related though.

madxime:
The IC bike SE will stop sending notifications if an unsupported Fitness Machine Control Point procedure is used. To prevent it, read the machine fitness features to determine if the Elapsed Time Supported and the Remaining Time Supported bits are set in order to determine if the "Start or Resume" can be used.

Also, reduce the connection parameters to 100 instead of 120 to be able to get the notifications which are sent every 500 ms.

This fixes #605

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .vscode/launch.json
 .vscode/ipch
 .vscode/extensions.json
+.idea/
 test/tmp_pio_test_transport.cpp
 data/config.txt
 include/telegram_token.h
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ SmartSpin_logfile.txt
 build/CMakeCache.txt
 *.log
 .DS_Store
+C/Users/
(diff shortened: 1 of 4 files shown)


doudar:
I appreciate your pull request! Just leaving a message here to let you know that this is important to me, and I will get around to testing it as soon as I can. Thanks for your contribution!

PR #626, opened 2025-02-12
Weighted downvoting

ddtomic:
NEEDS REVIEW

This PR is for implementing weighted downvoting, where neighbors are dynamically downvoted based on how far they are from the target value.

Issues: 

- Removing some of the magic numbers (table divisor) caused the power table to not build anymore
- neighbor target position is getting a value of 0

### DIFF
--- a/src/ERG_Mode.cpp
+++ b/src/ERG_Mode.cpp
@@ -85,7 +85,7 @@ void PowerBuffer::set(int i) {
   this->powerEntry[i].readings++;
   this->powerEntry[i].watts          = rtConfig->watts.getValue();
   this->powerEntry[i].cad            = rtConfig->cad.getValue();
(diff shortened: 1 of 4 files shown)


src/ERG_Mode.cpp
```diff
@@ -729,6 +729,39 @@ void PowerTable::clean() {
...
+  penalty = delta * penaltyFactor;
+
+  // quadratic function, high aggression
+  // penalty = penaltyFactor * (delta * delta)
+
+  // log function, medium aggression
+  // penalty = penaltyFactor * log(delta + 1);
+
+  // max penalty is 10
+  if (penalty > 10) {
```
doudar, line 754: Should this be MAX_NEIGHBOR_WEIGHT?

ddtomic: Correct, it should be, while debugging I made some changes and forgot to change that back, my apologies

doudar:
just setting this ready for review so I can merge it into my testing branch

PR #631, opened 2025-03-11
Update CSS loading mechanism

bannus:
The `loadCss()` code on a 100ms delay leads to a [Flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) on every page load in the web UI. Switch to the standard mechanism of loading a CSS file.

Feel free to reject if there's some technical reason the `loadCss()` code is required.


doudar:
I'll try to add this into the testing branch and see how it goes. With earlier (webserver library) versions, there was an issue where if the webserver received too many requests in a row, it would become unresponsive. This was a WiFi/BLE coexistence workaround.

doudar (approved):
merging into testing!

PR #632, opened 2025-03-13
Correct Cycling Power And CSC Feature Characteristics 

### LINKED ISSUE #630: [BUG] Invalid BLE value length (GTBike V crash)
### Description

When connecting to [GTBike V ](https://www.gtbikev.com) via BLE, the connection will make a few readings and then crash GTAV. Digging into the  GTBike V logs, this error message related to BLE appears:

```
System.ArgumentException: Destination array is not long enough to copy all the items in the collection. Check array index and length.
<...>
at Ble.Devices.Services.CPService.<Start>d__46.MoveNext()
```

Posting in the [GTBike V support forum](https://www.facebook.com/groups/1089053124812221/posts/2302052786845576/), the developer concluded this:

> Looks like your trainer is sending a non compliant value in the Cycling Power Feature characteristic. The BLE standard specifies that as a 2 byte value, but your trainer is sending a 1 byte value. When trying to convert that into a ushort, there is not enough data in the received array to make the conversion and it fails. I could make this a bit more robust, but it's very difficult to support hardware that does not comply with the standards.

The ask here is to **ensure that SS2K is following the BLE spec for these values.**

### Steps to reproduce

1. Install GTAV and GTBikeV 
2. Install the latest ScriptHookV and ScriptHookVDotnet
3. Have SmartSpin2k (v25.1.28) ready for BLE connection
4. Launch GTAV and enable the mod. The game will crash to desktop a few seconds later.

**Expected behavior**
The game should not crash and work, as it is claimed to be compatible [in the docs](https://docs.smartspin2k.com/compatibility.html#supported-training-programs).

**Actual behavior**
The game crashes.

### Additional Information
* Logs and config for my GTBikeV setup: https://gist.github.com/claabs/25f458cdc59a24d6b41ea4bbee0c6d68

matthewshirley:
It was reported that GTBikeV crashes when a SmartSpin2k device connects as the length of the Cycling Power feature characteristic is only 1-byte. As noted in the following [comment](https://github.com/doudar/SmartSpin2k/issues/630#issuecomment-2719457987), this characteristic should be 4-bytes.

Therefore, this PR updates the Cycling Power feature characteristic to be 4 bytes, with the 2nd and 3rd bit flagged. I additionally updated the Cycling Speed Cadence, which is supposed to be 2 bytes. However, the last byte is reserved, so I doubt any applications struggle with handling this.

PR resolves #630.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Fixed flags in CPS and CSC which were causing issues in GTA Bike. Thanks @matthewsshirley !
 
 ### Hardware
 
--- a/include/BLE_Cycling_Power_Service.h
+++ b/include/BLE_Cycling_Power_Service.h
@@ -21,4 +21,5 @@ class BLE_Cycling_Power_Service {
   BLECharacteristic *cyclingPowerMeasurementCharacteristic;
   BLECharacteristic *cyclingPowerFeatureCharacteristic;
   BLECharacteristic *sensorLocationCharacteristic;
+  byte cpFeature[4] = {0, 0, 0, 0};
 };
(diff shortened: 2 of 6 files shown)


doudar:
This looks pretty straightforward, with changes to only the flags. I tested with Zwift and TrainerRoad and both are working correctly. Thanks for your contribution!

doudar (approved):

PR #634, opened 2025-03-17
Full cubic spline implementation with manual calculations 

benv12:


doudar (approved):
Nice progress so far. Everything has been testing stable on a few test rides, so I'm going to approve this merge. We still need to complete a few finishing touches before we actually merge this into testing.

PR #710, opened 2025-12-30
Configurable inactivity timeout

claabs:
The 30 minute idle timeout reboot is mildly annoying as it causes my bike to beep on BLE reconnection. This makes the timeout configurable via the settings webpage. This way the reboot is still in place since it's necessary, but can just occur less often if a user desires it.

Feel free to change the order for where this setting lives on the settings page. I just put it at the bottom for now.

Since this is a new custom characteristic setting, it requires an app update to support it. Not sure what the release process is for that part.

----

P.S. During development I noticed my `settings.html` was not being updated from the official OTA update process; it was a version over 9 months old.  I had to use PlatformIO to erase and flash the filesystem manually. Not sure if this was just my device being weird.


doudar:
Thanks @claabs ! I'll start looking at this.

doudar:
testing before merge.

@claabs , have you found any issues with running longer reboots?

claabs:
I unfortunately haven't had time to test the longer reboot; I reverted to the develop branch since.

I can install this branch and get back to you in 48 hours if you like.

claabs:
I set the timeout to 48 hours and biked 2 sessions during it. No issues, and it made getting started quicker, since it didn't need to re-home!

PR #729, opened 2026-04-25
Fix Peloton serial homing range and post-homing starting gear

saffronconfidant9:
## Problem

Peloton serial users could not shift up in Zwift free-ride (sim) mode after the Jan 2 2026 commit that set gear 8 as the post-homing position. The root cause is two gaps in the Peloton serial homing path:

1. `pelotonConnected()` set the shift-blocking bounds (`setMinResistance` / `setMaxResistance`) but never called `resistance.setMin/setMax`, so `_findFTMSHome()` always saw `resistance.getMax() == 0` and skipped resistance-feedback homing entirely for serial Peloton users.

2. `goHome()` required `connectedPowerMeter != NONE` to take the resistance-feedback path, but serial Peloton has no BLE power meter. It fell through to StallGuard sensorless detection, which fires prematurely due to the Peloton's continuously-increasing magnetic resistance — leaving hMax at ~50% of the actual physical travel range.

Starting at gear 8 against a half-range hMax left no room to shift up.

## Fix (3 changes, 1 commit)

- **`pelotonConnected()`**: also call `resistance.setMin/setMax` so `_findFTMSHome()` receives valid `[5, 98]` resistance targets.
- **`goHome()`**: extend FTMS homing condition to include `ss2k->pelotonIsConnected`, routing serial Peloton through resistance-feedback homing instead of StallGuard.
- **`BLE_Client` startup**: replace hardcoded `setShifterPosition(8)` with `hMax / (2 * shiftStep)`, placing the motor at the center of the actual calibrated travel range regardless of bike type.

## Hardware tested

Validated on a physical Peloton with SmartSpin2k connected via serial (not BLE). After flashing and a Zwift spindown calibration:
- hMax corrected from ~12,698 to ~21,638 steps (full physical travel range)
- Full up and down shifting confirmed in Zwift free-ride mode
- Spindown calibration persists the correct hMax to config

## Notes

- The two pre-existing native test failures (`test_parses_cadence`, `test_fill_incomplete_table`) are present on the current `develop` tip and are unrelated to these changes.
- This fix was developed with AI assistance (Claude Sonnet 4.6, Anthropic). The `Co-Authored-By` trailer in the commit reflects this. All changes were reviewed and validated on hardware before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Fix Peloton serial homing to use resistance feedback instead of StallGuard sensorless detection.
+  StallGuard fired prematurely on Peloton's continuously-increasing magnetic resistance, causing
+  hMax to be set at ~50% of the actual physical travel range. pelotonConnected() now sets
+  resistance.setMin/setMax (the homing target bounds) so that _findFTMSHome() receives valid
+  targets, and goHome() now includes ss2k->pelotonIsConnected in its FTMS homing condition so
+  Peloton serial users take the resistance-feedback path rather than falling through to StallGuard.
+- Replace hardcoded post-homing gear 8 with a calculated midpoint (hMax / (2 * shiftStep)).
+  The hardcoded value assumed a correctly-ranged hMax; for Peloton serial users whose hMax was
+  only ~50% of physical range, starting at gear 8 left no room to shift up. The calculated
+  value places the motor at the center of the actual calibrated travel range.
 
 ### Hardware
 
--- a/src/BLE_Client.cpp
+++ b/src/BLE_Client.cpp
@@ -239,7 +239,8 @@ void bleClientTask(void* pvParameters) {
           ss2k->goHome(true);
         } else {  // Startup Homing
           ss2k->goHome(false);
-          rtConfig->setShifterPosition(8);  // Set to middle position after homing on startup
+          int startGear = (userConfig->getShiftStep() > 0) ? (userConfig->getHMax() / (2 * userConfig->getShiftStep())) : 4;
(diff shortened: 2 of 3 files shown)


doudar:
I believe the root cause here is that your bike wasn't traveling the full distance to the hardware stops during homing. I've made some changes in a different branch and merged them which should fix this issue, but I do plan to eventually merge a couple pieces of code from this PR. 

Mainly I'll probably implement resistance homing instead of sensor less for a serial connected Peloton. I was hesitant to do this in the past because the 1 and 100 resistance floats around a little bit in relation to the actual position of the magnetic brake as well as sensor less being a little bit faster. 

I was concerned about this causing incorrect power tables, but I suppose it really doesn't matter because if we're serial connected without a BLE power meter, we're using bike reported power anyway, so it's better if it matches resistance. 

Thanks for the PR!

saffronconfidant9:
Thank you for taking the time to review my submission! You are correct, what I had observed was that the ss2k would not travel to the full hardware/stepper motor-based stops when it would perform the homing. I only have the gen1 peloton bike to test with and am not familiar with what side effects this might have for others.

Since flashing the build from the PR branch I've noticed that after pedaling for ~2 secs when jumping on the bike fresh it will auto-home and seems to hit the 5 + 95 low/high resistance stops I have in the settings. That coupled with adjusting shift step size has made this excellent for using with Zwift. (for me)

This was my first couple weeks with the new bike and playing with the smartspin. Really appreciate all the work everyone has done, it's already made a drastic change for fitness in our household.

Also the first time I felt motivated/confident enough to submit a PR. I tried to be very focused in my use of LLMs, if there's anything I can do differently in the future I'd like to adjust. I don't know what I can commit to but would like to help burn down open issues, time permitting.

doudar:
Glad you appreciate it!

I'm always open to PR's wherever people see places to improve!

PR #745, opened 2026-06-25
Spinning Bike Chrono Support

eMadman:
Bike uses custom characteristics.  

- Device name: "CHRONO" + a number (mine shows as "CHRONO 12")
- Telemetry characteristic: A026E01D-0A7D-4AB3-97FA-F1500F9FEB8B (under service A026EE07-0A7D-4AB3-97FA-F1500F9FEB8B)
- Streams a 19-byte packet about once per second as soon as notifications are enabled, no handshake needed
- Packet starts with a fixed 3-byte header: 0x53 0x59 0x16
- Power: bytes 17-18, uint16 little-endian, watts, no scaling
- Cadence: bytes 8-9, uint16 little-endian ÷ 10, RPM
- Also available if useful: speed at bytes 6-7 ÷ 100 (km/h), distance at bytes 10-12, calories at bytes 13-14, heart rate at byte 5, elapsed time at bytes 3-4

User shared the following example implementation:

https://github.com/TrackMyIndoorWorkout/TrackMyIndoorWorkout/blob/develop/lib/devices/device_descriptors/precor_spinner_chrono_power.dart

I tried to follow the custom implementations used for other similar bikes (Echelon)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ C/Users/
 dependencies.lock
 .vscode/settings.json
 *.map
+graphify-out/
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -122,4 +122,8 @@
     "idf.pythonInstallPath": "C:\\Users\\anthony\\.espressif\\tools\\idf-python\\3.11.2\\python.exe",
     "idf.flashType": "UART",
     "liveServer.settings.port": 5501,
+    "clangd.arguments": [
+        "--compile-commands-dir=d:/git/SmartSpin2k/.cache/clangd",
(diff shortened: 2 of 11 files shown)


eMadman:
Feedback from the user:

> I tested the firmware today and am super happy to report it connects and values are correct with the built in display. So amazing work!
> 
> There is only one small thing that needs to be fixed - the bike gets discovered when clicking on scan and shows connected on the bikes computer screen, but on the smartspin ipad app it doesn't show it in the list. So in screenshot you actually see that there is no Chrono 12 item in the list, but by having Any selected it connects to bike and works as it should.

image
image

doudar (approved):

eMadman:
Updated html - pending user feedback before merge.

eMadman:
User confirmed web interface works correctly.  Ready to merge pending approval.

PR #762, opened 2026-09-14
Added variable gear ratios and web elements for selection 

doudar:


eMadman:
2 issues found and addressed:

1. Homing stopped on reboot due to  fantom shift
2. in unlimited gearing mode, it was possible to click the shifter beyond the upper endstop, requiring several down clicks before the knob would turn again.

